PR #2, opened 2016-06-26
Feature/docker native

jwtodd:
leverage native docker : http://beta.docker.com

note: more to do (eg remove brew install docker-machine, brew install native docker)


jwtodd:
@longdogz i have been able to start wasabi under both docker-machine and docker-native with this patch, eg:

```
# docker-machine
% ./bin/wasabi.sh -b true bootstrap start
# docker-native
# ensure docker-native is started/running
% OS=native ./bin/wasabi.sh -b true bootstrap start
```

note: we should likely add a note to the readme and/or anything else that surfaces as a result of this PR review.

longdogz:
FYI, @jwtodd, I fixed some problems with building and running the UI when using docker native.  Now if you run a command like this:

OS=native ./bin/wasabi.sh -b true start

And then do:

OS=native ./bin/wasabi.sh resource:ui

You will start the UI up.

ATTRIBUTIONS/wasabi-attribution.xml
```diff
@@ -611,4 +611,4 @@
          </downloadUrls>
       </dependency>
    </dependencies>
-</attributionReport>
+</attributionReport>
```
tariqislam: Any idea what that character at the end of the file/line is?

ATTRIBUTIONS/wasabi-experiment-objects-attribution.xml
```diff
@@ -535,7 +535,7 @@
          <type>jar</type>
          <licenses/>
          <downloadUrls>
-            <string>TBD</string>
+            <string>file:/Users/jtodd1/Documents/projects/wasabi/modules/experiment-objects/../../repository/Hyrule/Hyrule/0.0.1/Hyrule-0.0.1.jar</string>
```
tariqislam: Should this string be something else other than a local file?

bin/container.sh
```diff
@@ -129,14 +125,13 @@ stop_container() {
...
-
 remove_container() {
   [ ${1} ] && container=${1}
 
   if [ ${container} ]; then
     stop_container ${container} >/dev/null 2>&1
     docker rm -fv ${container} >/dev/null 2>&1
-  else
+  elif [ "$OS" == "OSX" ]; then
     docker-machine rm -f ${project} >/dev/null 2>&1
```
tariqislam: Question here: Would we ever want to remove the entire machine, since that will remove any cached images in that machine? Is it good enough to just remove the container?

bin/container.sh
```diff
@@ -237,9 +229,10 @@ console_mysql() {
 }
 
 status() {
-  eval $(docker-machine env ${project}) 2>/dev/null
-  docker-machine active 2>/dev/null | grep ${project} || usage "start ${project}" 1
-  eval $(docker-machine env ${project})
+  if [ "$OS" == "OSX" ]; then
+    docker-machine active 2>/dev/null | grep ${project} || usage "start ${project}" 1
```
tariqislam: In a new shell, without running the "eval $(docker-machine ....)" statement first, it seems like "docker-machine active" command returns "no active host found", hence the "usage ..." conditional statement is always run.

ATTRIBUTIONS/wasabi-experiment-objects-attribution.xml
wasabi-honeybadger: ahhhhh ... we need to run [ws]/bin/ATTRIBUTIONS_patch.sh

this is a stop-gap measure until we remove the local file repository, necessary now until we release hyrule source code.

i just pushed up a revision post filter.

bin/container.sh
wasabi-honeybadger: yeah that is intentional.

that said, when docker-native becomes the only supported impl this second conditional goes away all together ... thereby addressing the concern, at least as i read it.

bin/container.sh
wasabi-honeybadger: ok ... we can add the eval back in.

there is an always-invoked eval earlier on in the call stack and as such i had thought this was redundant. but if it is needed, we can add it back in.

tariqislam: Ah, I was unit testing instead of functional testing that section... Sounds good, thanks!

tariqislam:
👍

longdogz:
:shipit:

PR #18, opened 2016-06-29
add statistics integration test to testng suite

felixgao:
I have ran the test locally,  all test passes locally.

screen shot 2016-06-29 at 2 47 11 pm


modules/main/src/main/env/service/entrypoint.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
```
iizrailevsky: @felixgao Is this a new script? When would it be run?

modules/functional-test/testng_integrationTests.xml
```diff
@@ -78,6 +78,11 @@
             <class name="com.intuit.wasabi.tests.service.segmentation.MutualExclusionBatchTest"/>
         </classes>
     </test>
+    <test name="IntegrationStatisticsTests">
+        <classes>
+            <class name="com.intuit.wasabi.tests.service.statistic.StatisticTest"/>
```
iizrailevsky: @felixgao Awesome! I don't see the actual TestNG test class though. Not a part of this pull request? :)

modules/main/src/main/env/service/entrypoint.sh
wasabi-honeybadger: @felixgao  this looks to replicate ./bin/wasabi.sh test

if so, can we bring the concerns/objecties together?

modules/functional-test/testng_integrationTests.xml
felixgao: It is magic :) 
j/k
the testng test class is already committed before we pushed to wasabi

modules/main/src/main/env/service/entrypoint.sh
felixgao: This is the entrypoint.sh that will be used by the wasabi docker image for developer who want to build docker image for the current version of source code.  I will check in the dockerfile into a different branch but you want it in this branch that is fine with me.

felixgao: @wasabi-honeybadger  regarding looks to replicate wasabi.sh, it is meant to be similar with the exception of 
```code
while ! nc -w 1 -z wasabi-cassandra 9042; do sleep 0.1; done
while ! nc -w 1 -z wasabi-mysql 3306; do sleep 0.1; done
```
which checks for the database to be up before running the java code.  The reason to do that is java code sometimes runs a lot faster than the time it requires to bring up a database.  So when the database is not ready, our java webapp should not start.  It is just a safety precaution.

felixgao: for completeness, https://github.com/intuit/wasabi/pull/31 is where the dockerfile is that uses this script

iizrailevsky: @felixgao Got it - so entrypoint.sh is a run script for a docker image. Thanks

modules/functional-test/testng_integrationTests.xml
iizrailevsky: 👍

modules/main/src/main/env/service/entrypoint.sh
iizrailevsky: @felixgao Mind moving the modules/main/src/main/env/service/entrypoint.sh change to https://github.com/intuit/wasabi/pull/31 then so it's in one place?

iizrailevsky:
@felixgao :shipit: on adding the StatisticTest & let's move modules/main/src/main/env/service/entrypoint.sh change to https://github.com/intuit/wasabi/pull/31 as they're separate concerns

felixgao:
@iizrailevsky  done

iizrailevsky:
:shipit:

Issue #15, opened 2016-06-29
support linux


jwtodd:
**goal:** driver script should be able to support
- [ ] ubuntu:apt
- [ ] centos:yum
- [x] osx:homebrew

PR #23, opened 2016-06-30
add OSX check for package

shaikhjamir:
While executing the package target added OS check

### DIFF
--- a/bin/wasabi.sh
+++ b/bin/wasabi.sh
@@ -189,7 +189,9 @@ status() {
 }
 
 package() {
-  profile=build
+
+  profile_default_package=build
+  profile=${profile:=${profile_default_package}}
 
   ./bin/build.sh -b true -p ${profile}
 
--- a/bin/wasabi.sh
+++ b/bin/wasabi.sh
@@ -189,10 +189,19 @@ package() {
 
   ./bin/build.sh -b true -p ${profile}
 
-  (export VAGRANT_CWD=./bin; vagrant up)
-  (export VAGRANT_CWD=./bin; vagrant ssh -c "cd wasabi; mvn dependency:resolve")
+  if [ -z "$OS" ]; then
+    (export VAGRANT_CWD=./bin; vagrant up)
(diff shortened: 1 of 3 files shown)


bin/wasabi.sh
```diff
@@ -213,12 +222,14 @@ package() {
     done; \
     sed -i '' -e "s|http://localhost:8080|${server}|g" target/constants.json 2>/dev/null; \
     sed -i '' -e "s|VERSIONLOC|${version}|g" target/app/index.html 2>/dev/null; \
+    if [ -z "$OS" ]; then \
     (cd target; \
       npm install; \
       bower install; \
       grunt clean; \
       grunt build --target=develop --no-color); \
```
jwtodd: @longdogz in the context of jenkins, we still need to exec the above grunt build, no?

the objective here is to leverage ./bin/wasabi.sh to the fullest extent possible, which works quite well on osx but for linux we will assume (for now) the dependent toolchain we brew install today (eg vagrant, fpm, grunt, et al) is pre-installed, as it is in our internal system. as such, i suspect we need to exec the above non-install operations.

jwtodd:
+1 from me ... but let's here what @longdogz has to say

jwtodd:
**see:** #15

bin/wasabi.sh
longdogz: Sorry, @jwtodd , but I'm not sure what the context is.  Are you asking what we should do related to the UI when doing a jenkins build internally?  I guess it depends on whether you want to build and install the UI?  If so, then you need those commands.  It seems that "-z "$OS"" is a check if that var is empty, right?  I don't see where it gets set in the script.  But basically, those commands are necessary if you want the UI built at all.

Also, you mention something about grunt being pre-installed.  I don't think it is pre-installed on any system, certainly, not on our laptops.  It is on jenkins, because it was installed globally by hand (by me, probably).  So, again, not sure what you're asking.

wasabi-honeybadger:
@shaikhjamir blocked on this one? if so, let's go ahead with the merge.

wasabi-honeybadger:
@shaikhjamir :shipit:

PR #31, opened 2016-06-30
add dockerfile to be used for local development purpose

felixgao:
this docker file will allow user to run the current build in docker

### DIFF
--- a/modules/main/Dockerfile
+++ b/modules/main/Dockerfile
@@ -4,13 +4,14 @@ ENV WASABI_HOME /usr/local/jabba
 ENV WASABI_LOG_DIR /var/log/jabba
 
 
-ADD target/wasabi.main-*-all.jar ${WASABI_HOME}/wasabi.main.jar
 ADD target/classes/logback.xml ${WASABI_HOME}/conf/
 ADD target/classes/metrics.properties ${WASABI_HOME}/conf/
 ADD target/classes/web.properties ${WASABI_HOME}/conf/
 ADD target/extra-resources/service/entrypoint.sh ${WASABI_HOME}/entrypoint.sh
 
 RUN chmod 775 ${WASABI_HOME}/entrypoint.sh && mkdir ${WASABI_LOG_DIR}
+
+ADD target/wasabi-main-*-all.jar ${WASABI_HOME}/wasabi.main.jar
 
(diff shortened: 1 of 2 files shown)


jwtodd:
@felixgao the current toolchain supports this today, no?

that is not to say there is room for improvements though.

as such, can we bring these concerns together? vs introduction of unnecessarily disparent scripts/tooling/processes?

felixgao:
There are couple things the current toolchain does not confirm to conventions. 

- it uses centos base image which is about 400mb by it self. Adding java it is about 600mb without wasabi

- it add ./ instead of just the jar file and confs. (docker file suppose to be self documenting)

- the current entrypoint does not check for downstream system is listening to the port that we expects to be open

- It is anti pattern to have Dockerfile to be stored outside of the root directory of that project (sub-module)

- the current version also bundles the UI, which is not the direction we want to take for separating out the ui with its backend.  

I think this version is an incremental improvement.

iizrailevsky:
@felixgao @jwtodd @calvinwma Sounds like we want to discuss how to evolve tooling from developer's perspective: requirements & our approach. I'll set something up.

jwtodd:
@felixgao all things we can consider collaborating against the current codebase, correct?

i do not see a technical reason why this is not solvable.

not sure i totally agree/understand all the concerns re anti-patterns/etc ... but we have been on a journey to get where we are, there is **much** work to do but at the same time nothing need remain static per se. 

i am likely not alone in desiring that we keep in check drift of all sorts, be it code, tests, sh, docker, etc.

modules/main/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM felixgao/jre8-jacoco:8u92-0.7.6
+
+ENV WASABI_HOME /usr/local/jabba
```
shoeffner: Contains jabba in env variables!

Please take a look at our current Dockerfile and [feature/windows](https://github.com/intuit/wasabi/blob/1b83047b7b3c81aa3de313521b2b0de24ab6f9f7/modules/main/src/main/env/docker/wasabi/Dockerfile), maybe we can find a nice middleground?

felixgao: thanks, I will use your docker images. will close the pull request.

PR #33, opened 2016-07-02
First step to improve test quality

utobi:
A lot of unnecessary tests removed that are for example asserting not null on Object methods, asserting not null on methods that could be checking equality on actual values, or checking for equality on the same object.


iizrailevsky:
@utobi Thanks for pointing these out! Completely agree that a lot of these seem to be bogus checks. These need to be changed to compare with real values, as opposed to just "assertNotNull(...)". Removing them completely seems to decrease code coverage by 0.9% (see above).

AndreaSuckro:
These issues have been addressed with: https://github.com/intuit/wasabi/pull/123

PR #34, opened 2016-07-05
Adding .gitignore file to hello-world demo app

longdogz:
We want to ignore the node_modules directory (don't check those files in).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,6 @@ modules/ui/results
 modules/ui/coverage2
 modules/ui/results2
 modules/ui/dist
-modules/ui/node_modules
+node_modules
 modules/ui/app/bower_components
 modules/ui/app/scripts/config.js
--- a/contrib/hello-world/.gitignore
+++ b/contrib/hello-world/.gitignore
@@ -1,0 +1,1 @@
+node_modules
--- a/contrib/hello-world/.gitignore
+++ b/contrib/hello-world/.gitignore
@@ -1,1 +1,0 @@
-node_modules


wasabi-honeybadger:
@longdogz should we remove the entry in the top-level .gitignore that likely does the same thing as this patch? that line being:

    modules/ui/node_modules

tariqislam:
@wasabi-honeybadger I believe the top-level node_modules refers to the directory within modules, while the change here refers to node_modules directory created within the client_demo directory; hence the top-level one is needed.

tariqislam:
👍

longdogz:
No, @jwtodd .  That line is specific to the node_modules beneath modules/ui , right?  The one in the contrib/hello-world/.gitignore is specific to the node_modules in that contrib app.  I suppose we could change the one in the top level .gitignore to ignore ALL node_modules directories and the files beneath them, but I don't know the syntax for that.

wasabi-honeybadger:
ahhh right ... good point @longdogz ... your call

wasabi-honeybadger:
i wonder why travis went red on this push :(

jwtodd:
@longdogz ahhh this is more like what i meant to say:

  how about we consider adding a single-line entry in the top-level .gitgnore as follows: node_modules

this way, any/all directories named 'node_modules' will be ignored.

this is identical to the 'target' entry where every java module includes, post compile time, a 'target' directory we want to exclude.

we want to see a green travis build as well ... not sure what the issue is w/ that build just yet.

longdogz:
OK, I've made that change @jwtodd @tariqislam .

Issue #32, opened 2016-07-01
scrub legacy names


jwtodd:
**goal:** legacy name/package/etc need to be scrubbed

**see:** feature/scrub-legacy

- [X] [ws]/contrib/client-demo/error.html
- [x] [ws]/contrib/hello-world/app/index.html
- [x] [ws]/modules/api/src/test/java/com/intuit/wasabi/api/ExperimentsResourceTest.java
- [x] [ws]/modules/eventlog/README.md
- [x] [ws]/modules/functional-test/README.md

jwtodd:
@iizrailevsky @longdogz can you look into scrubbing '[jJ]abba' from the above listed contrib files?

it was obvious to me as to how to go about this.

i took on the other listed files.

@krushit i assume the above list captures the matches you found?

krushit:
$ egrep -r -i "abntest|jabba" .
./contrib/client-demo/error.html:	<script type="text/javascript" src="js/jabba-test-app/app.js"></script>
./contrib/client-demo/error.html:	<script type="text/javascript" src="js/jabba-test-app/controllers/mainController.js"></script>
./contrib/hello-world/app/index.html:

longdogz:
OK, I took care of the hello-world problem.  The client-demo one is not code I know, and looks like it might be referencing non-existent files.  Can you look at it @iizrailevsky @tariqislam ?

tariqislam:
@longdogz @krushit I'm looking at the client-demo ones.

jwtodd:
Resolved

PR #35, opened 2016-07-06
Feature/scrub legacy

jwtodd:
@krushit are you good with this?

see: #32


krushit:
:shipit: Looks good @jwtodd, I am merging it now

PR #40, opened 2016-07-12
contrib/client-demo:  Displaying http response codes for both API responses

jon077:
No associated issue.  Please advise if an issue should be created.

- Adding the http response codes to the API response
- For the acton API, required using "complete" instead of "success" ajax handler because the response content-type was causing jQuery to try and parse the empty body.

Consider creating an issue to update the action API to not include the content-type (which jQuery uses to try and parse the response in to JSON)

### DIFF
--- a/contrib/client-demo/js/wasabiTest.js
+++ b/contrib/client-demo/js/wasabiTest.js
@@ -43,8 +43,8 @@ $(document)
 																+ expLabel
 																+ "/users/"
 																+ userID,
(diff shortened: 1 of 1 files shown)


iizrailevsky:
Nice find and fix @jon077! 👍  Let's merge it. :shipit:

tariqislam:
👍

jwtodd:
awesome!

PR #47, opened 2016-07-18
Feature/rollup integration test

mans4singh:
This is the pull request for migrating python roll up tests to tests based on TestNG framework.


modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RollUpIntegrationTest.java
```diff
@@ -0,0 +1,1144 @@
...
+			}
+		}
+
+	}
+
+	@Test(dependsOnMethods = { "t_PostImpressionTomorrowQA" })
+	public void t_CheckImpressionForTomorrowByBucketFromMySQLProdContext()
+			throws Exception {
+
+		Statement statement = connection.createStatement();
```
Faedrivin: Here you can also use try-with-resource - this Statement and the ResultSet below are never closed.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RollUpIntegrationTest.java
```diff
@@ -0,0 +1,1144 @@
...
+				"creationTime", "modificationTime", "ruleJson");
+		experiment.setSerializationStrategy(experimentComparisonStrategy);
+
+	}
+
+	@BeforeClass
+	public void prepareDBConnection() throws Exception {
+		Class.forName("com.mysql.jdbc.Driver").newInstance();
+		connection = DriverManager.getConnection(
+				appProperties.getProperty("database.url"), "readwrite",
```
Faedrivin: Readwrite is also retrievable from the properties in this PR (see properties file below).

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RollUpIntegrationTest.java
```diff
@@ -0,0 +1,1144 @@
...
+		paramsQA.context = QA;
+
+		long startTime = System.currentTimeMillis();
+		ExperimentCounts countsBeforeQA = postExperimentCounts(experiment, paramsQA);
+		long endTime = System.currentTimeMillis();
+
+		System.out.println("Before analytics time QA was: "
+				+ (endTime - startTime));
+
+		startTime = System.currentTimeMillis();
```
Faedrivin: 1) Why do we need to measure the times here?
2) This one instance of creating the start time should be a little later, directly before postExperimentCounts is called, to make it more comparable. Also consider using [Instant or nanoTime](http://stackoverflow.com/a/3383047/3004221).

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RollUpIntegrationTest.java
```diff
@@ -0,0 +1,1144 @@
...
+				insertIntoRollUp(rollUpQuery);
+			}
+		}
+	}
+
+	private void insertIntoRollUp(String rollUpQuery) throws SQLException {
+		Statement statement = connection.createStatement();
+
+		ResultSet resultSet = statement.executeQuery(rollUpQuery);
+
```
Faedrivin: Better have a try-with-resource here - if there are problems in the resultSet, the statement might not close this way.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RollUpIntegrationTest.java
```diff
@@ -0,0 +1,1144 @@
...
+		Statement statement = connection.createStatement();
+		ResultSet result = statement.executeQuery(query);
+
+		if (!result.next())
+			fail("Should have at least one row for query " + query);
+
+		int userCount = result.getInt("user_count");
+		int distinctUserCount = result.getInt("distinct_user_count");
+
+		result.close();
```
Faedrivin: here as well: better use a try-with-resource to wrap the closes into a finally.

modules/functional-test/pom.xml
```diff
@@ -142,5 +142,10 @@
             <artifactId>retrofit1-okhttp3-client</artifactId>
             <version>1.0.2</version>
         </dependency>
+		<dependency>
+		    <groupId>mysql</groupId>
+		    <artifactId>mysql-connector-java</artifactId>
+		    <version>5.1.6</version>
+		</dependency>
```
Faedrivin: I just annotate it here, but it applies to multiple files: In general we indent with four spaces. One of the files below has 8, one has 2. This one has four, but the dependency should by one level closer to the root.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RollUpIntegrationTest.java
mans4singh: Will correct it.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RollUpIntegrationTest.java
mans4singh: You are right.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RollUpIntegrationTest.java
mans4singh: 1. To get an estimate of the processing time before hand.
2. Will correct it.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RollUpIntegrationTest.java
mans4singh: Will correct it.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RollUpIntegrationTest.java
mans4singh: Will correct it.

modules/functional-test/pom.xml
mans4singh: Will format it.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RollUpIntegrationTest.java
```diff
@@ -0,0 +1,1144 @@
...
+import com.jayway.restassured.response.Response;
+
+import static com.intuit.wasabi.tests.library.util.Constants.*;
+
+/**
+ * RollUp integration tests
+ */
+public class RollUpIntegrationTest extends TestBase {
+
+	private static final String CONTEXT = "context";
```
Faedrivin: This class seems to be using an indentation of 7 spaces.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RollUpBadCallsIntegrationTest.java
```diff
@@ -0,0 +1,254 @@
...
+import com.intuit.wasabi.tests.model.factory.UserFactory;
+import com.jayway.restassured.response.Response;
+
+import static com.intuit.wasabi.tests.library.util.Constants.*;
+/**
+ * Bucket integration tests
+ */
+public class RollUpBadCallsIntegrationTest extends TestBase {
+
+	private static final String FROM_TIME = "fromTime";
```
Faedrivin: Here the indentation is 7 spaces, further below correct with 4 but even further with only 2.

Faedrivin:
Apart from some indentations which can be resolved during the merge this is great, so :shipit:

mans4singh:
Formatted tests

PR #50, opened 2016-07-21
Changing so that hello world expects localhost for the local wasabi s…

longdogz:
…erver, rather than 192.xxx .  That is because we are shifting to using docker native instead of docker-machine.

### DIFF
--- a/contrib/hello-world/README.md
+++ b/contrib/hello-world/README.md
@@ -8,7 +8,7 @@ npm
 
 ##Configuration
 
-If you are not running a copy of Wasabi locally on your system at 192.168.99.100:8080, you will need
+If you are not running a copy of Wasabi locally on your system at localhost:8080, you will need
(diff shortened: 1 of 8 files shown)


tariqislam:
Question: Do we need to change the container.sh script or build.sh script (where it does a 'sed' on the index.html page)?

contrib/hello-world/README.md
```diff
@@ -8,7 +8,7 @@ npm
 
 ##Configuration
 
-If you are not running a copy of Wasabi locally on your system at 192.168.99.100:8080, you will need
+If you are not running a copy of Wasabi locally on your system at localhost:8080, you will need
 to search for "WASABI.setOptions" in the source (currently two locations) and then change the server
```
Faedrivin: I would rather name the locations or leave out the "currently two locations" - Also maybe a hint for searching for localhost might be smarter: You don't mention Gruntfile.js or the shell scripts which one would need to change as well.

Issue #45, opened 2016-07-15
Failed tests:   testRunFinally(com.intuit.wasabi.eventlog.impl.EventLogImplTest): (..)


wasabi-honeybadger:

shoeffner:
resolved in #61

Issue #22, opened 2016-06-29
testSystem(com.intuit.wasabi.eventlog.EventLogSystemTest) : sporadic failure


jwtodd:
**goal:** ELSTest.testSystem() flaps

**status:** @Ignore ELSTest

shoeffner:
resolved in #61

Issue #20, opened 2016-06-29
testPostEvent(com.intuit.wasabi.eventlog.impl.EventLogImplTest): failing intermittently


iizrailevsky:
testPostEvent(com.intuit.wasabi.eventlog.impl.EventLogImplTest)  Time elapsed: 0.623 sec  <<< FAILURE!
    Wanted but not invoked:
    eventLogListener.postEvent(
        com.intuit.wasabi.eventlog.events.SimpleEvent@2c95ac9e
    );
    -> at com.intuit.wasabi.eventlog.impl.EventLogImplTest.testPostEvent(EventLogImplTest.java:210)
    Actually, there were zero interactions with this mock.
    	at com.intuit.wasabi.eventlog.impl.EventLogImplTest.testPostEvent(EventLogImplTest.java:210)
    	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    	at java.lang.reflect.Method.invoke(Method.java:483)
    	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
    	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
    	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
    	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
    	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
    	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
    	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
    	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
    	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
    	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
    	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
    	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
[...]

PR #61, opened 2016-08-11
simplifying eventlog tests, trying to remove thread reliance

shoeffner:
Removed lots of parts from the EventLog tests to make them slimmer and to attempt solving #20 #22 #45. We might need to dive into PowerMock to mock Thread.sleep in testPostEvent and testRunFinally, but I do hope that 5 seconds of optional timeout are enough for now.

### DIFF
--- a/modules/eventlog/src/test/java/com/intuit/wasabi/eventlog/impl/EventLogImplTest.java
+++ b/modules/eventlog/src/test/java/com/intuit/wasabi/eventlog/impl/EventLogImplTest.java
@@ -15,6 +15,7 @@
  *******************************************************************************/
 package com.intuit.wasabi.eventlog.impl;
 
+import com.intuit.wasabi.eventlog.EventLog;
 import com.intuit.wasabi.eventlog.EventLogListener;
 import com.intuit.wasabi.eventlog.events.EventLogEvent;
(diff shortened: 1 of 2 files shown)


Faedrivin:
-> remove one test which was basically there for coverage (only calling the code, not actually checking results), hence the reduction by 0.02.

modules/eventlog/src/test/java/com/intuit/wasabi/eventlog/impl/EventLogImplTest.java
```diff
@@ -24,10 +25,8 @@
 import org.mockito.Mockito;
 
 import java.lang.reflect.Field;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
iizrailevsky: Please specify classes used, instead of .*

iizrailevsky:
@shoeffner Looks good! Apart from a small class import fix, shipit! 👍 :shipit:

modules/eventlog/src/test/java/com/intuit/wasabi/eventlog/impl/EventLogImplTest.java
shoeffner: Good catch!

PR #64, opened 2016-08-16
Removing 192.* from README.md since docker native is now being used.

tariqislam:


README.md
```diff
@@ -107,15 +101,32 @@ Congratulations! You are the proud owner of a newly minted Wasabi instance. :)
...
 #### Troubleshooting
 
-* While starting Wasabi, if you run into errors such as this, run the following command in
-your terminal and re-run ./bin/wasabi.sh start:
+* While starting Wasabi, if you see an error when the docker containers are starting up, you could do the following:
+
+** Look at the current docker containers that have been successfully started.
+
+```bash
+% docker ps
```
iizrailevsky: @tariqislam Instead of dealing with docker commands directly (which they could if they really want to), they can actually do "./bin/wasabi.sh status" - which does the same thing. Should we point to this instead?

iizrailevsky:
@tariqislam With exception of s|docker ps|./bin/wasabi.sh status| looks good! 👍

README.md
tariqislam: Yes, since we have it, let's point to status instead.

PR #66, opened 2016-08-17
Changing default ip to localhost since docker native is now used.

tariqislam:


iizrailevsky:
@tariqislam Awesome! 👍 :shipit:

PR #69, opened 2016-08-18
removing python dependency in shell scripts

shoeffner:
removes the python scripts by leveraging help:evaluate better


jwtodd:
@shoeffner awesome! will look into this shortly.

jwtodd:
@shoeffner this is awesome! not sure how i missed this feature when doing this first implementation. and one less package dependency is significant.

:shipit:

PR #72, opened 2016-08-24
Adding Hello World demo app to show how to use Wasabi for AB testing …

longdogz:
…within a React app.


contrib/hello-world-react/.eslintrc
```diff
@@ -0,0 +1,30 @@
+{
+	"env": {
+		"browser": true,
+		"node": true,
+		"es6": true,
+		"jest": true,
+		"commonjs": true
+	},
+	"extends": "eslint:recommended",
```
shoeffner: I think it is better to extend from `eslint:recommended` and `plugin:react/recommended` which basically makes [line 28](https://github.com/intuit/wasabi/pull/72/files#diff-42f4433a10d03d8372dbd9078124bb1dR28) obsolete and allows for Components to be named whatever you want.

contrib/hello-world-react/.htmlhintrc
```diff
@@ -0,0 +1,23 @@
...
+  "attr-value-double-quotes": true,
+  "attr-value-not-empty": false,
+  "attr-no-duplication": true,
+  "doctype-first": false,
+  "tag-pair": true,
+  "tag-self-close": false,
+  "spec-char-escape": true,
+  "id-unique": true,
+  "src-not-empty": true,
+	"title-require": true,
```
shoeffner: nit: indentation

contrib/hello-world-react/src/server.js
```diff
@@ -0,0 +1,43 @@
+import path from 'path';
+import http from 'http';
+import express from 'express';
+import bodyParser from 'body-parser';
+
+export default function(config) {
+
+	const app = express();
+	const server = http.createServer(app);
```
shoeffner: nit: indentation

contrib/hello-world-react/src/www/css/styles.scss
```diff
@@ -0,0 +1,83 @@
+$text-color: black;
+
+body {
+	color: $text-color;
```
shoeffner: nit: indentation

contrib/hello-world-react/src/server.js
```diff
@@ -0,0 +1,43 @@
...
+
+    let products = [
+        {id: '1', name: 'iphone 6', cost: 300, quantity: 0},
```
shoeffner: The quantities here should be obsolete by the changes we discussed earlier, right? You should never be accessing/changing them anymore.

contrib/hello-world-react/src/www/js/components/item-form.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
```
shoeffner: You don't use this file at all.

contrib/hello-world-react/src/www/js/components/login.js
```diff
@@ -0,0 +1,147 @@
...
+            errorMessage: ''
+        };
+
+        this.onChange = this.onChange.bind(this);
+        this.validate = this.validate.bind(this);
+        this.submit = this.submit.bind(this);
+    }
+
+    componentDidMount() {
+        // let session = sessionStorage.getItem('session');
```
shoeffner: I guess this method can simply be removed.

contrib/hello-world-react/src/www/js/components/store-list.js
```diff
@@ -0,0 +1,225 @@
...
+        WASABI.setOptions({
+            'applicationName': 'MyStore',
+            'experimentName': 'TestBuyButtonColor',
+            'protocol': 'http',
+            'host': 'localhost:8080'
+        });
+
+        var session = sessionStorage.getItem('session');
+        if (session) {
+            this.state.session = JSON.parse(session);
```
shoeffner: Maybe you should add session to the state initialization in l. 58

contrib/hello-world-react/src/www/js/components/store-list.js
```diff
@@ -0,0 +1,225 @@
...
+        {
+            'userID': this.state.session.login.name
+        }).then(
+            function (response) {
+                console.log('postAction: success');
+                if (response) {
+                    console.log(JSON.stringify(response));
+                }
+            },
+            function (error) {
```
shoeffner: it was discouraged to use .then(cb,ecb) but instead .then(cb).catch(ecb)

shoeffner:
Additionally to what is pointed out above, I found that in multiple files there are tabs, in others spaces, in some files it gets mixed. You might want to unify that.

Apart from those few things it works well!

shoeffner:
I just realized that's it's possible to buy negative quantities (i.e. having them in your cart). But of course that doesn't really matter for this demo.

shoeffner:
Here is a beautiful squirrel for you: 

A Squirrel

:shipit:

PR #73, opened 2016-08-25
Fixed discrepancy of naming email properties in pom.xml,

jcwuzoegiver:
email.properties and EmailModule. Also removed intuit reference of
email.host.


jcwuzoegiver:
@iizrailevsky @jwtodd Please review and bless it.

ATTRIBUTIONS/wasabi-email-attribution.xml
```diff
@@ -1214,7 +1214,7 @@
          <type>jar</type>
          <licenses/>
          <downloadUrls>
-            <string>TBD</string>
+            <string>file:/Users/jwu/git/wasabi-core/modules/email/../../repository/Hyrule/Hyrule/0.0.1/Hyrule-0.0.1.jar</string>
```
shoeffner: This can't be right either :-(

modules/email/pom.xml
```diff
@@ -32,8 +32,8 @@
     <properties>
         <email.service.enabled>true</email.service.enabled>
         <email.service.class.name>com.intuit.wasabi.email.impl.NoopEmailImpl</email.service.class.name>
-        <email.host>mail.intuit.com</email.host>
-        <email.from>no-reply-wasabi@intuit.com</email.from>
+        <email.service.host>localhost</email.service.host>
+        <email.service.from>no-reply-wasabi@intuit.com</email.service.from>
```
shoeffner: I am not sure if this should be @intuit.com, maybe we should better use example.com?

jcwuzoegiver: agreed @shoeffner. ill fix the email.service.from too.

ATTRIBUTIONS/wasabi-email-attribution.xml
jcwuzoegiver: @shoeffner good catch. @jwtodd, any idea how this string was generated from attribution maven plugin?

jwtodd: yes ... this is an unfortunate artifact of having a project internal mvn repo for hyrule, which we will fix.

till then, run:

`  % ./bin/ATTRIBUTIONS_patch.sh`

then git add, etc.

ATTRIBUTIONS/wasabi-functional-test-attribution.xml
```diff
@@ -825,6 +825,23 @@
...
+         <projectUrl>http://dev.mysql.com/doc/connector-j/en/</projectUrl>
+         <type>jar</type>
+         <licenses>
+            <license>
+               <name>The GNU General Public License, Version 2</name>
+               <url>http://www.gnu.org/licenses/old-licenses/gpl-2.0.html</url>
+            </license>
+         </licenses>
+         <downloadUrls>
+            <string>https://repo.maven.apache.org/maven2/mysql/mysql-connector-java/5.1.6/mysql-connector-java-5.1.6.jar</string>
```
jwtodd: hmmm, we have to reg this dep as such? maven central can't just server up the file dep via the conventional means?

modules/email/pom.xml
```diff
@@ -32,8 +32,8 @@
     <properties>
         <email.service.enabled>true</email.service.enabled>
         <email.service.class.name>com.intuit.wasabi.email.impl.NoopEmailImpl</email.service.class.name>
-        <email.host>mail.intuit.com</email.host>
-        <email.from>no-reply-wasabi@intuit.com</email.from>
+        <email.service.host>localhost</email.service.host>
+        <email.service.from>no-reply-wasabi@example.com</email.service.from>
```
jwtodd: +1 ... dooooo this :)

jwtodd:
:shipit:

PR #74, opened 2016-08-29
'bin/wasabi.sh test:module-name,another-module' will run unit tests f…

shoeffner:
…or those modules and only compile what's necessary for it

### DIFF
--- a/bin/wasabi.sh
+++ b/bin/wasabi.sh
@@ -282,6 +282,8 @@ for command in ${@:$OPTIND}; do
     start) command="start:cassandra,mysql,wasabi";&
     start:*) commands=$(echo ${command} | cut -d ':' -f 2)
       (IFS=','; for command in ${commands}; do start ${command}; done);;
+    test:*) commands=$(echo ${command} | cut -d ':' -f 2)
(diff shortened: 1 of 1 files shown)


felixgao:
does the shell script works within the submodules?

bin/wasabi.sh
```diff
@@ -282,6 +282,8 @@ for command in ${@:$OPTIND}; do
     start) command="start:cassandra,mysql,wasabi";&
     start:*) commands=$(echo ${command} | cut -d ':' -f 2)
       (IFS=','; for command in ${commands}; do start ${command}; done);;
+    test:*) commands=$(echo ${command} | cut -d ':' -f 2)
```
felixgao: also seems you need to add that change to the prompt for test subcommand

shoeffner:
You just call it as you always do: from root call bin/wasabi.sh test:experiment-objects for example. It won't work when invoked from modules/experiment-objects via ../../bin/wasabi.sh - if I understand your question correctly.

bin/wasabi.sh
shoeffner: Added it to the usage with [7092306](https://github.com/intuit/wasabi/pull/74/commits/7092306ea567b3866c8652cf9e5b6f2cf071565f)

felixgao:
:shipit:

jwtodd:
color me : impressed @shoeffner  :)

:shipit:

PR #75, opened 2016-08-29
Feature/jwtodd

jwtodd:
'wasabis.sh package' builds app and ui rpm and deb files

### DIFF
--- a/bin/wasabi.sh
+++ b/bin/wasabi.sh
@@ -167,7 +167,7 @@ resource() {
 }
 
 stop() {
-  ./bin/container.sh stop${1:+:$1}
+  ./bin/congit tainer.sh stop${1:+:$1}
 }
 
 status() {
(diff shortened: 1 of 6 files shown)


jwtodd:
@vinodtalati let's do this :)

bin/wasabi.sh
```diff
@@ -173,7 +182,7 @@ resource() {
 }
 
 stop() {
-  ./bin/container.sh stop${1:+:$1}
+  ./bin/congit tainer.sh stop${1:+:$1}
```
vinodtalati: @jwtodd  should this be container.sh?

vinodtalati:
Hi James this looks good. Please see my comment for the possible typo. Other than that it looks good. Lets worry about the merge with my changes after this is through.

bin/wasabi.sh
jwtodd: ouch yes ... how did that happen ... good call

jwtodd:
@vinodtalati merged ... let's get your PR merge ready :)

i will help as needed.

PR #78, opened 2016-09-08
Feature/windows

shoeffner:
- Supplying several scripts for windows to build/start/stop/remove wasabi via docker-machine.
- Minimizing Dockerfile to only 6 steps.
- Adjusted (i.e. simplified) container.sh according to the learnings I had with windows.
- Added instructions to `README.md`.
- Changing IntegrationTest behavior slightly: fixing some tests (rollups, tests using assignments) and duct-taped the bucket tests.


bin/container.sh
```diff
@@ -141,27 +141,18 @@ start_wasabi() {
...
-  wmip=$(docker inspect --format "{{ .NetworkSettings.Networks.${docker_network}.IPAddress }}" ${project}-mysql)
 
   remove_container ${project}-main
 
   if [ "${verify}" = true ] || ! [ docker inspect ${project}-main >/dev/null 2>&1 ]; then
     echo "${green}${project}: building${reset}"
 
-#    sed -i -e "s|\(http://\)localhost\(:8080\)|\1${mip}\2|g" modules/main/target/${id}/content/ui/dist/scripts/config.js 2>/dev/null;
-    docker build -t ${project}-main:${USER}-$(date +%s) -t ${project}-main:latest modules/main/target/${id}
+    docker build -t ${project}-main:$(git rev-parse --short=8 HEAD) -t ${project}-main:latest modules/main/target/${id}
```
jwtodd: nice :)

bin/container.sh
```diff
@@ -141,27 +141,18 @@ start_wasabi() {
...
   echo "${green}${project}: starting${reset}"
 
-  wenv="WASABI_CONFIGURATION=-DnodeHosts=${wcip} -Ddatabase.url.host=${wmip}"
-
-#   fixme: try to reuse the start_container() method instead of 'docker run...' directly; currently a problem with quotes in ${wenv} being passed into container.
-  docker run --net=${docker_network} --name ${project}-main -p 8080:8080 -p 8090:8090 -p 8180:8180 \
-    -e "${wenv}" -d ${project}-main || \
-    usage "docker run --net=${docker_network} --name ${project}-main -p 8080:8080 -p 8090:8090 -p 8180:8180 -e \"${wenv}\" -d ${project}-main" 1
-
+  start_container ${project}-main ${project}-main:$(git rev-parse --short=8 HEAD) "-p 8080:8080 -p 8090:8090 -p 8180:8180 -e WASABI_CONFIGURATION=\"-DnodeHosts=${project}-cassandra -Ddatabase.url.host=${project}-mysql\""
```
jwtodd: quite nice :)

bin/container.sh
```diff
@@ -175,39 +166,24 @@ start_cassandra() {
 }
 
 console_cassandra() {
-  wcip=$(docker inspect --format "{{ .NetworkSettings.Networks.${docker_network}.IPAddress }}" ${project}-cassandra)
-
-  docker run --net=${docker_network} -it --rm ${cassandra} cqlsh ${wcip} || \
-    usage "unable to run command: docker run --net=${docker_network} -it --rm ${cassandra} cqlsh ${wcip}" 1
+  docker exec -it ${project}-cassandra cqlsh || \
```
jwtodd: my thoughts around using 'run' vs 'exec' was that it ensured a "new" cqlsh client could network connect to the cas container, vs just starting a cqlsh client on the existing cas container ... which doesn't seem like a real network-reachability test. make sense?

thoughts?

bin/container.sh
```diff
@@ -175,39 +166,24 @@ start_cassandra() {
...
-    create database if not exists ${project};
-    grant all privileges on ${project}.* to 'readwrite'@'localhost' identified by 'readwrite';
-    grant all on *.* to 'readwrite'@'%' identified by 'readwrite';
-    flush privileges;
-EOF
-)
-
-  docker run --net=${docker_network} -it --rm ${mysql} mysql -h${wmip} -P3306 -uroot -p${pwd} -e "${sql}" || \
-    usage "unable to run command: % docker run --net=${docker_network} -it --rm ${mysql} mysql -h${wmip} -P3306 -uroot -p${pwd} -e \"${sql}\"" 1
+  start_container ${project}-mysql ${mysql} "-p 3306:3306 -e MYSQL_ROOT_PASSWORD=${pwd} -e MYSQL_DATABASE=wasabi -e MYSQL_USER=readwrite -e MYSQL_PASSWORD=readwrite"
```
jwtodd: really nice :) ... assuming the grants apply intrinsically?

nit: s/wasabi/${project}/

bin/container.sh
```diff
@@ -175,39 +166,24 @@ start_cassandra() {
...
   [ "${verify}" = true ] && console_mysql
 }
 
 console_mysql() {
   pwd=mypass
-  wmip=$(docker inspect --format "{{ .NetworkSettings.Networks.${docker_network}.IPAddress }}" ${project}-mysql)
 
-  docker run --net=${docker_network} -it --rm ${mysql} mysql -h${wmip} -P3306 -uroot -p${pwd} || \
-    usage "unable to run command: % docker run --net=${docker_network} -it --rm ${mysql} mysql -h${wmip} -P3306 -uroot -p${pwd}" 1
+  docker exec -it ${project}-mysql mysql -uroot -p${pwd} || \
```
jwtodd: my thoughts around using 'run' vs 'exec' was that it ensured a "new" cqlsh client could network connect to the cas container, vs just starting a cqlsh client on the existing cas container ... which doesn't seem like a real network-reachability test. make sense?

thoughts?

bin/wasabi.sh
```diff
@@ -231,7 +231,7 @@ resource() {
         open http://localhost:8080/swagger/index.html;;
       doc) [ ! -f ./target/site/apidocs/index.html ] && build
         open ./target/site/apidocs/index.html;;
-      mysql|cassandra) ./bin/wasabi.sh status 2>/dev/null | grep wasabi-${1} 1>/dev/null || ./bin/wasabi.sh start
+      mysql|cassandra) ./bin/wasabi.sh status 2>/dev/null | grep wasabi-${1} 1>/dev/null || ./bin/wasabi.sh start:${1}
```
jwtodd: great catch :)

bin/wasabi.bat
```diff
@@ -0,0 +1,122 @@
+@echo off
```
jwtodd: what is this thing ?windows? you speak of?

modules/main/src/main/env/docker/wasabi/Dockerfile
```diff
@@ -13,34 +13,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ###############################################################################
-from centos
```
jwtodd: purdy

jwtodd:
@shoeffner commented against code ... i will run it locally rsn. nicely done :)

bin/container.sh
shoeffner: @felixgao's idea :-)

bin/container.sh
shoeffner: > MYSQL_USER, MYSQL_PASSWORD
These variables are optional, used in conjunction to create a new user and to set that user's password. This user will be granted superuser permissions (see above) for the database specified by the MYSQL_DATABASE variable. Both variables are required for a user to be created.
[Environment Variables Docker MySQL](https://hub.docker.com/_/mysql/)

I thought the database is not configured by maven, but will adjust that.

bin/container.sh
shoeffner: I see, makes sense. On windows I just used `exec` because it just worked and wondered why we would introduce the additional complexity. I will try to simplify `run` then so that we don't need the `${wcip}` but can use `${project}-cassandra` instead as well.

bin/container.sh
shoeffner: see above

modules/ui/Gruntfile.js
```diff
@@ -7,6 +8,15 @@
 // use this if you want to recursively match all subfolders:
 // 'test/spec/**/*.js'
 
+
+// Allows to read a parameter from the command line argument list.
+// If a user wants to set 'key' she should call `grunt [commands...] --key=value`.
+// If the key is not supplied, the default value is used.
+function fromArgs( key, defaultValue ) {
+	const index = process.argv.findIndex(arg => arg.startsWith('--' + key + '='));
```
shoeffner: needs proper indentation

modules/main/src/main/env/service/run
```diff
@@ -48,7 +48,10 @@ if [ "$JAVA_HOME" = "" ]; then
     export JAVA_HOME=$(/usr/libexec/java_home -v 1.8)
   elif [ -d /usr/java/latest ]; then
     # Linux
-    export JAVA_HOME=/usr/java/latest  
+    export JAVA_HOME=/usr/java/latest
+  elif [ -f /usr/bin/java ]; then
+    # alpine/Docker
+	export JAVA_HOME=/usr
```
shoeffner: indentation

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RollUpIntegrationTest.java
```diff
@@ -138,6 +138,9 @@ public RollUpIntegrationTest() throws Exception {
     @BeforeClass
     public void prepareDBConnection() throws Exception {
         Class.forName("com.mysql.jdbc.Driver").newInstance();
+        setPropertyFromSystemProperty("database.url", "database.url");
```
shoeffner: move to TestBase for future portability

bin/win/bootstrap.bat
```diff
@@ -0,0 +1,115 @@
...
+if not %errorLevel% == 0 (
+    call :error You need administrator rights to bootstrap Wasabi.
+    endlocal
+    exit /b 1
+)
+
+rem install chocolatey
+call :debug Trying to find Chocolatey
+if not exist C:\ProgramData\chocolatey\choco.exe (
+  call :info Installing Chocolatey
```
shoeffner: indentation

README.md
```diff
@@ -390,3 +446,33 @@ Steps to contribute:
...
+Wasabi to connect to cassandra and MySQL inside the docker network.
+
+**How to connect the UI to IDE's wasabi-process?**
+You will also need to redirect the UI. By default `bin\wasabi.bat resource:ui`
+points to `docker-machine ip wasabi`. You can use the Windows exclusive
+`bin\wasabi.bat resource:dev_ui` to use `localhost` as your UI's target.
+
+**Wasabi has incorrect times! My IDE-run integration tests fail! Help!**
+It is a known problem with docker-machine that its time server stops while
+the host machine sleeps. If your system behaves weird check if
```
shoeffner: weird, check

README.md
```diff
@@ -77,13 +77,71 @@ Bootstrapping on Ubuntu requires sudo privileges to install all the required dep
...
-Similar tooling will work for Windows. Contribute a patch :)
+
+##### Windows (7+)
+
+To install Wasabi's dependencies on Windows we use [Chocolatey][win_choco] which
+needs administrator rights in your [cmd.exe][win_cmd_admin].
+
+If you have git, just run:
+```dos
+git clone https://github.com/intuit/wasabi.git
```
shoeffner: maybe streamline with unix/osx format of $ in front? I just used this without much thinking as I like easily copyable text...

jwtodd:
works for me:

    % ./bin/wasabi.sh bootstrap build start test stop

bin/win/bootstrap.bat
```diff
@@ -0,0 +1,115 @@
...
+
+call :info Installing/Upgrading Node.JS dependencies
+set remaining_packages=%npm_packages%
+:inst_npm_deps
+    for /f "tokens=1* delims=," %%P in ("%remaining_packages%") do (
+        set current_package=%%P
+        set remaining_packages=%%Q
+
+        call :debug Installing/Upgrading via npm - !current_package!
+        cmd /c "C:\Program Files\nodejs\npm.cmd" -g install !current_package!
```
AndreaSuckro: I am not too professional with all the scripting magic, but are there windows specific variables for paths like 'C:\Program Files\' ,'C:\ProgramData'(see chocolatey installation)? Something like CSIDL_PROGRAM_FILES that could be used here to avoid problems on different Windows machines aka different localized names? Will try to run the files on my German machine this evening and let you know ;)

shoeffner: Wherever you got CSIDL from, I used %ProgramData% and %ProgramFiles% now. However, for Ruby it seems that I have to rely on C:\Tools... or on a restart and using the "in path" variant.

AndreaSuckro: I got them from here: https://technet.microsoft.com/en-us/library/cc749104(v=ws.10).aspx - but some of them are not working on my system- who knows what those are :D Yours are working though.

bin/win/bootstrap.bat
```diff
@@ -0,0 +1,115 @@
...
+net session >nul 2>&1
+if not %errorLevel% == 0 (
+    call :error You need administrator rights to bootstrap Wasabi.
+    endlocal
+    exit /b 1
+)
+
+rem install chocolatey
+call :debug Trying to find Chocolatey
+if not exist C:\ProgramData\chocolatey\choco.exe (
```
AndreaSuckro: %ProgramData% here aswell?

shoeffner: It's fixed in the latest commit but it seems the reviews don't reflect those changes anymore or something... don't know. If you have time please check it out again and try to run it again.

alanjahns:
Some notes from my experience using this with Windows 7 Enterprise:
- The UI needs a separate command to run at all, would be nice if Wasabi would serve it as it does for OSX or at least explicitly state the extra steps when in the Windows Env "bin\wasabi.bat resource:ui"
- The download-way does not work as the build process needs an initialized git repository. I see "fatal: Not a git repository (or any of the parent directories): .git"
- A Windows machine may need virtualization enabled in the system bios, probably varies by manufacturer whether or not that is a default setting. Was not for my Lenovo W530.

shoeffner:
@alanjahns I addressed the issues.
- For now I just added notes about the UI to the README.md.
- The bootstrap process will initialize a git repository and reset to develop. (Careful while testing, as you will end up on that branch instead of feature/windows.)
- Added notes about virtualization in the FAQ.

modules/main/src/main/env/docker/wasabi/Dockerfile
```diff
@@ -13,34 +13,12 @@
...
+FROM openjdk:8-jre-alpine
 
-ENV WASABI_MODULE wasabi-main
-ENV WASABI_UI_MODULE wasabi-ui-main
-ENV WASABI_PORT 8080
-ENV WASABI_JMX_PORT 8090
-ENV WASABI_DEBUG_PORT 8180
+COPY ./ /usr/local/${application.name}
```
felixgao: Let's add a todo on making these magic files that appeared in docker more standard by adding from explicit path

bin/wasabi.bat
```diff
@@ -0,0 +1,122 @@
...
+                     set parameters=%%Q
+                    call :debug Remaining parameters: !parameters!
+                    
+                    rem Append parameter to parameter list - now space separated
+                    set space_parameters=!space_parameters! !param!
+                    call :debug Collected parameters: !space_parameters!
+                )
+                if defined parameters goto :read_parameters
+            
+            set assembled_command=bin/win/!command_name!.bat!space_parameters!
```
shoeffner: These should be backslashes - will need to check through all files.

shoeffner:
@alanjahns I made the UI blocking when calling `resource:ui` or `resource:dev_ui`. Is it working this way?

shoeffner:
@alanjahns I updated the README to only include mentions of the zip file, not specific instructions. I hope that works better. Can you please take a moment and review so we can finally get this in?

alanjahns:
@shoeffner Looks good. Thanks!

alanjahns:
No objections on my part @AndreaSuckro

Issue #56, opened 2016-08-04
contrib a simple single-node terraform deploy script


jwtodd:
goal: add a simple aws terraform deployment script

- [ ] assume package (eg wasabi.rpm) is built local
- [ ] instantiate single ec2 node, reasonably locked down and durable storage not required
- [ ] yum install: java, cassandra, mysql (may not need runit)
- [ ] rpm install wasabi
- [ ] start wasabi
- [ ] open external port
- [ ] test via % ./bin/wasab.ish -e IP:8080 test
- [ ] destroy instance
- [ ] commit as a contrib project

shoeffner:
Closing this, #80 was merged.

PR #80, opened 2016-09-12
terraform for single node wasabi setup

aqua0ps:
#56


shoeffner:
Once Travis is done we can ship it, it works good!

PR #84, opened 2016-09-19
reverted pagination for /applications/{appName}/experiments as it oth…

shoeffner:
…erwise will break customer code

### DIFF
--- a/modules/api/src/main/java/com/intuit/wasabi/api/ApplicationsResource.java
+++ b/modules/api/src/main/java/com/intuit/wasabi/api/ApplicationsResource.java
@@ -69,19 +69,17 @@ public class ApplicationsResource {
     private Authorization authorization;
     private Pages pages;
     private Priorities priorities;
-    private PaginationHelper<Experiment> experimentPaginationHelper;
 
     @Inject
     ApplicationsResource(final AuthorizedExperimentGetter authorizedExperimentGetter, final Experiments experiments,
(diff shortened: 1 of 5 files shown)


shoeffner:
Oh wow, the build failed? I have to look into that, it did not locally o.O

AndreaSuckro:
Looks good to me (although it is a bit sad). I removed the experimentPaginationHelper as well, but you are free to change that back if you think it not necessary. Otherwise :shipit:

shoeffner:
Can you check the integration tests once more, @AndreaSuckro ? Just want to be sure. :-)

AndreaSuckro:
Ok I think I found something exciting. When debugging the issue on my local machine I noticed that the IntegrationTests are only failing when wasabi was run in docker. The first test that is failing is checking whether Assignments can be generated. It fails, because a check on the current date is done and compared with the startTime of the experiment. 

So I did the following:
```
docker exec -it wasabi-main bash
date
Fri Oct  7 09:51:22 UTC 2016
```
The date in the docker machine is somehow completely off. I am not sure why this happens or if this is something happening only on my system, but I am pretty sure that this causes all the troubles I had with running the tests.

shoeffner:
Do you use docker machine? I have a fix for that on the windows branch, but thought native docker would be unaffected...

AndreaSuckro:
Ok - after removing docker completely it worked! Sorry for the delay. Please :shipit:

PR #90, opened 2016-09-26
Removed operators and data types (date) not supported by the version …

longdogz:
…of the rule parser used by Wasabi.  Also added online help to describe the supported features.


modules/ui/app/scripts/services/RuleEditFactory.js
```diff
@@ -20,9 +19,7 @@ angular.module('wasabi.services').factory('RuleEditFactory', ['UtilitiesFactory'
                 { menuLabel: 'does not equal', stringForm: '!=', jsonForm: 'notEquals' },
                 { menuLabel: 'equals (same case)', stringForm: '^=', jsonForm: 'stringEqualsExact' },
```
shoeffner: Since we annotate this with "same case", it might be an idea to note that equals is case insensitive in the string case? I am not too sure about this, it's fine this way as well (plus has been this way for a long time).

modules/ui/app/styles/wasabi.scss
```diff
@@ -4227,6 +4227,7 @@ form input.ng-dirty.ng-invalid:not(.ng-focused) {
     ul {
         list-style-position: inside;
         list-style-type: disc;
+        margin-left: 10px;
```
shoeffner: Maybe the whitespace between the · and the text could be a little smaller (and you could probably even add some more margin here).

shoeffner:
Overall the changes look fine, I was only confused when testing it:
When I "test" a rule, I suddenly don't have to entered quoted strings anymore, what I had to do when entering strings for the rules. Maybe it's possible to add those strings automatically (and not showing them in the UI at all)?

modules/ui/app/scripts/services/RuleEditFactory.js
longdogz: If you're OK with it, I think we'll just keep it the way it is, although I'll see if Calvin has an opinion.

modules/ui/app/styles/wasabi.scss
longdogz: I adjusted this.

longdogz:
After talking with Calvin, I am adjusting the bullets, but not the other things you mentioned.  Thanks!

shoeffner:
Very good, :shipit:

PR #91, opened 2016-09-27
Feature/leroy jenkins

jwtodd:
this patch includes modifications necessary for the wasabi-internal process

@iizrailevsky @shoeffner


bin/build.sh
```diff
@@ -37,7 +37,7 @@ EOF
 }
 
 fromPom() {
-  mvn -f $1/pom.xml -P$2 help:evaluate -Dexpression=$3 -B \
+  mvn ${WASABI_MAVEN} -f $1/pom.xml -P$2 ${WASABI_MAVEN} help:evaluate -Dexpression=$3 -B \
```
shoeffner: Why twice `${WASABI_MAVEN}`?

bin/build.sh
```diff
@@ -130,13 +130,13 @@ if [[ "${build}" = true || "${build_js}" = true ]]; then
       sudo gem install compass
     fi
   fi
-  (cd ./modules/ui && npm install && bower install && grunt build)
+  (cd ./modules/ui && npm install && bower install && grunt --forcce build)
```
shoeffner: Double 'c' in `forcce`. Not sure if you need it at all. Also why is the need to add /target/ ?

bin/wasabi.sh
```diff
@@ -247,7 +249,7 @@ status() {
 }
 
 package() {
-  profile=build
+  [ "${profile}" == "${profile_default}" ] && profile=build
```
shoeffner: So we will never be able to package development again?

bin/wasabi.sh
```diff
@@ -272,10 +274,11 @@ package() {
     done; \
     sed -i '' -e "s|http://localhost:8080|${server}|g" target/constants.json 2>/dev/null; \
     sed -i '' -e "s|VERSIONLOC|${version}|g" target/app/index.html 2>/dev/null; \
-    if [[ "${WASABI_OS}" == "${WASABI_OSX}" || "${WASABI_OS}" == "${WASABI_LINUX}" ]]; then \
+# SCOTT and SEBASTIAN
```
shoeffner: Remove unnecessary comments here.

modules/ui/bin/fpm.sh
```diff
@@ -51,22 +51,27 @@ id=${name}-${version}-${profile}
 home=${home:-/usr/local/$id}
 log=${log:-/var/log/$id}
 email=`fromPom main ${profile} application.email`
+# FIXME: hack
```
shoeffner: Why do we need this hack? What's not working?

bin/build.sh
jwtodd: great observation ... i opted to front the mvn invocation with the optional env-var overrides, yet neglected to remove the original ref.

bin/build.sh
jwtodd: thx ... was having trouble building dist w/o --force ... at one time.

but, given there is a typo ... i will give it a go w/o --force

re "/target/" can you ref a line?

in general, it is desirable/optimal to put gen'd files and such in a "target" dir so that it is not git added, cleaned up with 'mvn clean', etc. does this help?

bin/wasabi.sh
jwtodd: yeah that is a ramification.

this is needed to get build travis correctly (signing jars, etc) ... we could look at melding the profiles (build -> development) ... but we'd need to look into the ramifications, reasons it was separated out, etc.

honestly, i would like to revisit this ... but not block on this one at this point in time. as such, added a FIXME

bin/wasabi.sh
jwtodd: good call.

@longdogz 
note: there is a commented out trailing '... grunt test' ... do we need to include that?

modules/ui/bin/fpm.sh
jwtodd: wasn't able to pull that materialized value from the pom ... will revisit

bin/build.sh
```diff
@@ -78,10 +78,11 @@ module=main
 if [[ "${build}" = true || "${test}" = true || "${build_jar}" = true ]]; then
   [ "${build}" = true ] && package=package
   [ ! -e ./modules/main/target/wasabi-main-*-SNAPSHOT-${profile}-all.jar ] && package=package
-  [ "${test}" = true ] && tests=test
+#  [ "${test}" = true ] && tests="org.jacoco:jacoco-maven-plugin:prepare-agent cobertura:cobertura -Dcobertura.report.format=xml findbugs:check test"
```
felixgao: Remove dead code comment?

bin/build.sh
```diff
@@ -136,6 +137,7 @@ fi
 content=${home}/${id}/content/ui/dist
 
 mkdir -p ${content}
+#cp -R ./modules/ui/target/dist/ ${content}
```
felixgao: remove dead code comment

bin/wasabi.sh
```diff
@@ -216,7 +219,6 @@ resource() {
         ./bin/wasabi.sh status >/dev/null 2>&1 || ./bin/wasabi.sh start:docker
 #        jip=localhost
```
felixgao: remove this comment

bin/wasabi.sh
```diff
@@ -265,22 +268,17 @@ package() {
 
   ./bin/fpm.sh -n ${name} -v ${version} -p ${profile}
 
+# FIXME: don't rebuild, cp dist/* target/*
   (cd modules/ui; \
     mkdir -p target; \
     for f in app bower.json Gruntfile.js constants.json karma.conf.js karma-e2e.conf.js package.json test .bowerrc; do \
```
felixgao: why do we need test to be bundled?

jwtodd:
internal tests are now green ... getting ready to merge this PR @iizrailevsky

bin/build.sh
jwtodd: resolved

bin/build.sh
jwtodd: resolved

bin/wasabi.sh
jwtodd: resolved

bin/wasabi.sh
jwtodd: more details?

jwtodd:
@iizrailevsky @felixgao @shoeffner re-review : https://github.com/intuit/wasabi/pull/91

ATTRIBUTIONS/wasabi-analytics-attribution.xml
```diff
@@ -1158,7 +1158,7 @@
          <type>jar</type>
          <licenses/>
          <downloadUrls>
-            <string>TBD</string>
+            <string>file:/Users/jtodd1/Documents/projects/wasabi/modules/analytics/../../repository/Hyrule/Hyrule/0.0.1/Hyrule-0.0.1.jar</string>
```
shoeffner: please remove your path here

ATTRIBUTIONS/wasabi-cassandra-datastax-attribution.xml
```diff
@@ -0,0 +1,886 @@
+<attributionReport>
```
shoeffner: this file shouldn't be here (yet - it's on feature/datastax-driver-migration)

ATTRIBUTIONS/wasabi-repository-datastax-attribution.xml
```diff
@@ -0,0 +1,2282 @@
+<attributionReport>
```
shoeffner: This file is also irrelevant at the moment.

bin/wasabi.sh
```diff
@@ -27,6 +28,7 @@ reset=`tput sgr0`
 export WASABI_OS=${WASABI_OS:-`uname -s`}
 export WASABI_OSX="Darwin"
 export WASABI_LINUX="Linux"
+export WASABI_MAVEN=${WASABI_MAVEN}
```
shoeffner: This seems redundant, WASABI_MAVEN should already be an environment variable.

bin/wasabi.sh
shoeffner: There's a test directory (`package.json test .bowerrc`), I think @felixgao wants to know if it's necessary to bundle it, i.e. having it inside the target dir.

I assume we don't need it since we don't test the ui, but if we would, we might?

bin/wasabi.sh
```diff
@@ -265,22 +266,17 @@ package() {
 
   ./bin/fpm.sh -n ${name} -v ${version} -p ${profile}
 
+# FIXME: don't rebuild, cp dist/* target/*
```
shoeffner: `if [-d dist ]; then cp -r dist target; else ... fi`

bin/wasabi.sh
```diff
@@ -356,7 +356,7 @@ sleep=${sleep:=${sleep_default}}
 for command in ${@:$OPTIND}; do
   case "${command}" in
     bootstrap) bootstrap;;
-    build) build true;;
+    build) build true ${verify} ${profile};;
```
shoeffner: 👍

modules/database/src/main/resources/database.properties
```diff
@@ -13,10 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ###############################################################################
-database.url.host=${mysql.host}
-database.url.port=${mysql.port}
-database.url.dbname=${mysql.dbName}
-database.url.args=${mysql.Args}
+database.url.host:${mysql.host}
```
shoeffner: I think we usually use `=` in the other property files. I would vote going with that, as it allows to also execute them to set variables in a shell.

modules/email/src/test/java/com/intuit/wasabi/email/EmailModuleTest.java
```diff
@@ -20,9 +20,7 @@
 import com.google.inject.Injector;
 import org.junit.Test;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.fail;
+import static org.junit.Assert.*;
```
shoeffner: * import

modules/eventlog/src/main/java/com/intuit/wasabi/eventlog/impl/EventLogImpl.java
```diff
@@ -180,13 +196,27 @@ public void run() {
...
     /*test*/ void prepareEnvelope(final EventLogEvent event) {
-        if (event == null) {
-            return;
-        }
-        for (EventLogListener eventLogListener : listeners.keySet()) {
-            if (isSubscribed(eventLogListener, event)) {
-                eventPostThreadPoolExecutor.submit(new EventLogEventEnvelope(event, eventLogListener));
+        LOGGER.debug("preparing event: {}", event);
+
+        if (event != null) {
```
shoeffner: I prefer fail fast with shallow nesting. But I guess it does not matter.
Also I do not know if you want to have this many debug log statements here, it feels more like a trace level.

modules/main/src/main/resources/logback.xml
```diff
@@ -109,6 +109,31 @@
     </logger>
     -->
 
+    <logger name="com.intuit.wasabi.auditlog.impl.AuditLogEntryEnvelope" level="DEBUG">
```
shoeffner: Do you still need all these appenders?

modules/repository/src/main/java/com/intuit/wasabi/repository/impl/cassandra/CassandraAuditLogRepository.java
```diff
@@ -186,27 +196,39 @@ public boolean storeEntry(AuditLogEntry entry) {
...
         }
-        return true;
+
+        LOGGER.debug("stored auditLogEntry: {}, status: {}", entry, status);
+
+        return status;
     }
 
+    private String emptyStringIfNull(final String object) {
+        return object != null ? object : "";
```
shoeffner: I prefer `return Optional.ofNullable(object).orElse("");`, especially since @felixgao recommended avoiding `null` where possible.

modules/ui/bin/fpm.sh
shoeffner: did you revisit this/will it get fixed?

pom.xml
```diff
@@ -405,12 +388,22 @@ http://www.w3.org/2001/XMLSchema-instance">
                     <aggregate>true</aggregate>
                     <instrumentation>
                         <excludes>
-                            <exclude>com/intuit/wasabi/tests/**/*.class</exclude>
+                            <exclude>com/intuit/wasabi/**/*Test.class</exclude>
```
shoeffner: This probably won't exclude all functional tests. the /tests/** was there to exclude those.

pom.xml
```diff
@@ -405,12 +388,22 @@ http://www.w3.org/2001/XMLSchema-instance">
...
                     </instrumentation>
                 </configuration>
             </plugin>
             <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>findbugs-maven-plugin</artifactId>
+                <version>3.0.4</version>
+                <configuration>
+                    <xmlOutput>true</xmlOutput>
+                    <!--<findbugsXmlOutput>true</findbugsXmlOutput>-->
```
shoeffner: Unnecessary comment

ATTRIBUTIONS/wasabi-analytics-attribution.xml
jwtodd: resolved: had to run [ws]/bin/ATTRIBUTION_patch.sh

bin/wasabi.sh
jwtodd: ok good question. @longdogz ?

ATTRIBUTIONS/wasabi-cassandra-datastax-attribution.xml
jwtodd: good call ... thx

bin/wasabi.sh
jwtodd: concur

bin/wasabi.sh
jwtodd: i would like to defer fixing this one at this time ... it requires time to slow down and do it right. we always have subsequent patches, thus the FIXME.

modules/database/src/main/resources/database.properties
jwtodd: either way ... i tend to prefer ':' but that's a nit.

defer given we should fix this app-wide ... we can always issue subsequent patches.

modules/email/src/test/java/com/intuit/wasabi/email/EmailModuleTest.java
jwtodd: concern re the wildcard import?

this is default intellij ... which again i seldom-if-ever do anything w the imports directly.

defer for broader discussion

modules/eventlog/src/main/java/com/intuit/wasabi/eventlog/impl/EventLogImpl.java
jwtodd: i prefer one exit per method

it is a bit much of the logging ... but each one of those helped in this case. will revisit.

defer

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/PaginationTest.java
```diff
@@ -233,6 +234,7 @@ public void t_ExperimentPaginationFilterByStatePaused() {
     }
 
     @Test(dependsOnGroups = {"pagination_pages"}, groups = {"auditlog_smoke"})
+    @RetryTest(warmup = 5000)
```
jwtodd: good call

modules/main/src/main/resources/logback.xml
jwtodd: nope thx

modules/repository/src/main/java/com/intuit/wasabi/repository/impl/cassandra/CassandraAuditLogRepository.java
jwtodd: i like

pom.xml
jwtodd: concur

modules/main/pom.xml
```diff
@@ -82,7 +82,7 @@
             <id>test</id>
             <properties>
                 <!-- NOTE: do not pretty-print format -->
-                <application.instrument>-javaagent:/usr/local/lib/org.jacoco.agent-${jacoco.version}-runtime.jar=destfile=${LOG_DIR}/jacoco-it.exec,append=false</application.instrument>
+                <application.instrument>-javaagent:/usr/local/lib/org.jacoco.agent-0.7.2.201409121644-runtime.jar=destfile=${LOG_DIR}/jacoco-it.exec,append=false</application.instrument>
```
jwtodd: the jacoco version is hardwired like other deps and u/g to the latest.

modules/email/src/test/java/com/intuit/wasabi/email/EmailModuleTest.java
shoeffner: When I joined the project people always pointed it out, and I think it's nice to have it explicit rather than implicit - importing only specific parts is in general better than general parts. But it's fine, it works.

modules/main/pom.xml
shoeffner: Then why do we not hardwire wasabi... just sayin'

shoeffner:
:shipit: but keep the defered issues in mind.

Issue #88, opened 2016-09-23
creating an event (impression) for an unassigned user gives a 201 repsonse code


sudheerkethireddy:
I have created an experiment within an app. I tried creating an impression for a user without assigning the user to any bucket. The Api gives 201 as response code rather than throwing some warning/error saying that the user is not assigned.

Steps to re-produce.
1) create an App (if not existing)
2) create an experiment within the application.
3) create buckets 
4) do a Post call to submit an event (of type impression) and it returns 201.

PR #92, opened 2016-09-27
creating Event For Unassigned User

sudheerkethireddy:
Written automation to automate below bug 
https://github.com/intuit/wasabi/issues/88.


modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/assignment/CreateEventForUnassignedUserTest.java
```diff
@@ -0,0 +1,110 @@
...
+import com.intuit.wasabi.tests.library.TestBase;
+import com.intuit.wasabi.tests.library.util.serialstrategies.DefaultNameExclusionStrategy;
+import com.intuit.wasabi.tests.model.Assignment;
+import com.intuit.wasabi.tests.model.Experiment;
+import com.intuit.wasabi.tests.model.User;
+import com.intuit.wasabi.tests.model.factory.ExperimentFactory;
+import com.intuit.wasabi.tests.model.factory.UserFactory;
+
+/**
+ *
```
shoeffner: These doc comments (especially the `@created` one etc. can be left out - CVS takes care of those for us.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/assignment/CreateEventForUnassignedUserTest.java
```diff
@@ -0,0 +1,110 @@
+/*
+ * ******************************************************************************
+ *  * Copyright 2016 Intuit
```
shoeffner: Please double check how we usually include the license file, I think we just use one asterisk per line.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/assignment/CreateEventForUnassignedUserTest.java
```diff
@@ -0,0 +1,110 @@
...
+ * test: do a post on the
+ * aftertest: clean 
+ */
+public class CreateEventForUnassignedUserTest  extends TestBase{
+
+	static final Logger LOGGER = LoggerFactory.getLogger(CreateEventForUnassignedUserTest.class);
+	Experiment experiment;
+
+	@BeforeClass
+	public void t_setup()
```
shoeffner: Just call this `setup()`, we try to move all names to `camelCase` instead of `snake_case`.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/assignment/CreateEventForUnassignedUserTest.java
```diff
@@ -0,0 +1,110 @@
...
+	{
+		//create an experiment and populate the experiment POJO		
+		experiment = ExperimentFactory.createExperiment();
+		DefaultNameExclusionStrategy experimentComparisonStrategy = new DefaultNameExclusionStrategy("creationTime", "modificationTime", "ruleJson");
+		experiment.setSerializationStrategy(experimentComparisonStrategy);
+		Experiment exp = postExperiment(experiment);
+		Assert.assertNotNull(exp.creationTime, "Experiment creation failed (No creationTime).");
```
shoeffner: Instead of using all these asserts, you should be able to make the setup dependent on the Experiment-tests - or for the moment just rely on the fact that those things are tested elsewhere. That will save you a lot of code and we don't need duplicate testing.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/assignment/CreateEventForUnassignedUserTest.java
```diff
@@ -0,0 +1,110 @@
...
+		Assert.assertNotNull(exp.modificationTime, "Experiment creation failed (No modificationTime).");
+		Assert.assertNotNull(exp.state, "Experiment creation failed (No state).");
+		experiment.update(exp);
+
+		//create buckets within the experiment
+		String url = "/experiments/"+experiment.id+"/buckets";
+		String bucketA = "{\"label\": \"red\", \"allocationPercent\": 0.5, \"isControl\": false, " +
```
shoeffner: There's also a BucketFactory to avoid manual JSON creation:

    List<Bucket> buckets = postBuckets(BucketFactory.createCompleteBuckets(experiment, 2));

Will create a red and a green bucket and post them.
In general the return codes should be asserted by the TestBase class automatically - postBucket does so. So whenever possible don't use the apiServerConnector directly but refer to the TestBase classes instead.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/assignment/CreateEventForUnassignedUserTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+		//change the state of the experiment from DRAFT to RUNNING
+		assertReturnCode(response, HttpStatus.SC_CREATED);
+		response = apiServerConnector.doPut("experiments/"+experiment.id, "{\"state\": \"RUNNING\"}");
+		assertReturnCode(response, HttpStatus.SC_OK);
+	}
+
+
+	@Test()
+	public void t_replicateBug()
```
shoeffner: See above for the name (also check `t_cleanUp()` - just remove `t_` everywhere).
Also it would be better to give the test a more meaningful name. For example:

    testEventSubmissionBeforeAssignment()

shoeffner:
We usually use 4 spaces for indentation, you might want to change this here and update your editor settings.

Maybe you can move this test to the `EventsIntegrationTest`? It would seem to be a good addition there. If you think it is better to stand alone, how about adding it to the TestNG `testng_integrationTests.xml` file? Just add another class entry to the event or assignment section there.

And I just looked over all the other "assignment" related tests, we definitely need to clean up there 👿...

This is a nice find and a useful addition, next step for us is to get the test green!

@iizrailevsky We need to discuss how to handle these sorts of tests: if we integrate it properly, it will fail until fixed, if we don't integrate it, it might get forgotten - and this seems an important bug.

@sudheerkethireddy How does the database reflect this kind of event? Is it created, or is the event ignored entirely?

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/assignment/CreateEventForUnassignedUserTest.java
sudheerkethireddy: thanks shoeffner for the comment. i referenced the tests in the functional-test module for licesing and followed the same. Will change it as specifed above

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/assignment/CreateEventForUnassignedUserTest.java
sudheerkethireddy: Ok @shoeffner. I will make the changes accordingly.Thanks for pointing me to those util methods.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/assignment/CreateEventForUnassignedUserTest.java
sudheerkethireddy: Yeah i completely agree @shoeffner . I thought about it but used the snake_case so that the test i wrote is in sync with other tests. Will try to change them.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/assignment/CreateEventForUnassignedUserTest.java
evgavry: @sudheerkethireddy please use this class as an example of how the license file should look like: https://github.com/intuit/wasabi/blob/develop/modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/assignment/BasicAssignment.java

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/assignment/CreateEventForUnassignedUserTest.java
sudheerkethireddy: Thanks @shoeffner . Incorporated those utils and checked those changes.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/assignment/CreateEventForUnassignedUserTest.java
sudheerkethireddy: @evgavry  Thanks. Incorporated those changes.

evgavry:
Thank you @sudheerkethireddy if @shoeffner and other folks don't have any objections, I think it could be merged.  We can enable this test (add to testng_integrationTests.xml) when the bug is fixed.

shoeffner:
@sudheerkethireddy @evgavry , I discussed this with @iizrailevsky and he thinks we should not merge it until its fixed, so that we don't lose track of it. 
@bargenilesh @krushit , does anyone of you have time and is interested in tackling this one?

PR #97, opened 2016-10-03
Fix logback issue

jcwuzoegiver:
1) fileNamePattern was missing "${LOG}." and this created lots of *.tmp files and was causing disk to get full.
2) Replaced rollingPolicy with ch.qos.logback.core.rolling.FixedWindowRollingPolicy. This will prevent run out of disk space if load is really high for peak season.

@jwtodd


jwtodd:
:shipit:

PR #99, opened 2016-10-06
Fix to problem with displaying assignment counts when only one bucket

longdogz:
Initial changes to fix problem that if there is only one bucket, we weren't getting the count in the list. Fix is that the code to get the count into the objects used for creating the list was in a kind of loop that isn't run when there's only one.  Moved the code to a loop that is done even when there is only one.


shoeffner:
Looks good and works! :shipit:

Issue #100, opened 2016-10-06
Docker does not start after "wasabi.sh bootstrap" with Docker version upgraded


iizrailevsky:
I did "wasabi.sh bootstrap" which upgraded my Docker version. However Docker was NOT started automatically when running "wasabi.sh start" and I got "sleep" messages instead:

```
$ ./bin/wasabi.sh start
chill'ax 🍺 
chill'ax 🍺 
chill'ax 🍺 
chill'ax 🍺 
chill'ax 🍺 
chill'ax 🍺 
```

jwtodd:
see: feature/bug-osx-docker-start

iizrailevsky:
Fixed now! 👍  @jwtodd @shoeffner

PR #101, opened 2016-10-06
mod: s/wasabi_os_default/WASABI_OSX/g

jwtodd:
#100 

test:
- stop docker-native (note: can uninstall as well, which is what 'bootstrap' does)
- ./bin/wasabi.sh start

### DIFF
--- a/bin/container.sh
+++ b/bin/container.sh
@@ -70,7 +70,7 @@ beerMe() {
 
 start_docker() {
   docker ps >/dev/null 2>&1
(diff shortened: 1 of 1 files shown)


iizrailevsky:
@jwtodd [cc: @shoeffner] Just verified that it's starting properly on the feature/bug-osx-docker-start branch on my machine with Docker not started initially. Nice find @jwtodd @shoeffner! 👍  :shipit:

PR #104, opened 2016-10-07
fix the column name to sneak casing

felixgao:
This is the code to rename the camel casing column name to sneak casing.  I have ran unit test and functional-test which seems to be working.


shoeffner:
It looks fine and works with the UI :shipit:

PR #109, opened 2016-10-13
Addition of and support for Traffic Management plugins

longdogz:
This consists of extending the plugin mechanism of the Wasabi UI to support a top level Plugins tab.  Plugins can be configured that will contribute at that level.  Configured plugins will result in a description appearing on the Plugins tab and a Run button.  When the button is pressed, the UI contributed as the plugin is displayed in a dialog.

This pull request also includes some backend changes to add an API for getting daily actual percentages assigned to experiments.  This is used by one of the included plugins.

The included plugins are a Traffic Management UI and a Traffic Analysis UI.  The Traffic Management UI can be used when you have several mutually exclusive experiments and you want to achieve specific percentages of actual traffic for each.  The Traffic Analysis UI is useful for confirming that you are achieving those percentages.  The code for these plugins (and the configuration and a README) is in the contrib folder.


modules/api/src/test/java/com/intuit/wasabi/api/ExperimentsResourceTest.java
```diff
@@ -273,24 +274,24 @@ public void postExperiment() throws Exception {
...
+        } catch (AuthenticationException ignored) {
         }
 
         // When user(TESTUSER) doesn't have create permissions but flags is true
         // we create a new application and add him as ADMIN_LABEL
         when(authorization.getUser(AUTHHEADER)).thenReturn(TESTUSER);
         when(experiments.getExperiment(newExperiment.getID())).thenReturn(experiment1);
         Response responseNewApp = experimentsResource.postExperiment(newExperiment, true, AUTHHEADER);
-        assert(experiment1.equals(responseNewApp.getEntity()));
+        assert (experiment1.equals(responseNewApp.getEntity()));
```
AndreaSuckro: Wondering: why java assert instead of junit asserts?

modules/api/src/test/java/com/intuit/wasabi/api/ExperimentsResourceTest.java
```diff
@@ -1212,14 +1112,57 @@ public void experimentReturnsLocationHeader() throws Exception {
...
+        Experiment actualExperiment = experimentsResource.getAuthorizedExperimentOrThrow(experiment.getID(), USER);
+        Assert.assertEquals("Wrong experiment returned.", experiment, actualExperiment);
+
+        // no permission
+        doThrow(AuthenticationException.class).when(authorization).checkUserPermissions(USER, experiment.getApplicationName(), Permission.READ);
+        try {
+            experimentsResource.getAuthorizedExperimentOrThrow(experiment.getID(), USER);
+            Assert.fail("Should throw AuthenticationException if user has no permission.");
+        } catch (AuthenticationException e) {
+            e.printStackTrace();
```
AndreaSuckro: Is it necessary to print the stacktrace here?

contrib/traffic-management-plugins/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+```
+./bin/wasabi.sh stop:wasabi start:wasabi
+```
+
+(NOTE: If you are running the Wasabi UI locally, e.g., using "grunt serve", simply stop
+that command and run it again.)
+
+To confirm successful installation of the plugins, login to the UI and click on the
+Plugins tab.  You should see descriptions of the two plugins and buttons to launch them.
```
AndreaSuckro: I think this is just a minor thing - but since I started from scratch and had no experimentes, I first had to generate one to be able to click on the plugins tab.

contrib/traffic-management-plugins/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+##Usage
+
+When you have successfully installed the plugins and restarted the Wasabi server, you
+can login and click on the Plugins tab to use the plugins.  You will see descriptions
+of the two plugins and a button labeled Run next to each.  To use a plugin, simply
+click on the Run button and a dialog will appear with the UI for the plugin.
```
AndreaSuckro: If I click on the run button- nothing happens. Here is what I did: I am using wasabi from my IDE so after copying the 'plugins' folder over and replacing the plugins.js (I checked the paths they seem correct) I shut down the ui and did a ```grunt serve```. The plugins appear in the tab, but the button shows no response (and I also have a mutual exclusive experiment running). What am I doing wrong?

modules/api/src/test/java/com/intuit/wasabi/api/ExperimentsResourceTest.java
shoeffner: Because I didn't change it, the autoformatter just added the spaces there... Should I go through it and replace them?

modules/api/src/test/java/com/intuit/wasabi/api/ExperimentsResourceTest.java
shoeffner: It certainly is not, will fix it.

contrib/traffic-management-plugins/README.md
longdogz: This is the first step in the Installation section of the README for the plugins:

Simply copy the plugins directory (which contains 3 other directories) from this directory into the modules/ui/app directory.

Did you do that?  If you only moved the plugins.js file, then it is telling the UI that you have these plugins, and hence, the buttons for them get created and shown, but the code to actually implement the UI for them hasn't been put in a place where they can be loaded and run, e.g., in the modules/ui/app directory.  I guess we could have added more error handling in the plugin system?  But I hope that if you follow the instructions in the README, it will work better for you?

contrib/traffic-management-plugins/README.md
longdogz: That is how it always is.  Nobody (but someone testing the system) ever sees that, e.g., in Production.  If you have no experiments, we need to give you the default page.

longdogz:
BTW, thanks for doing this review, @AndreaSuckro !!!

contrib/traffic-management-plugins/README.md
AndreaSuckro: Ok my reading competence is at fault here I think - I copied the plugins (3 folders: traffic-analysis, traffic-management and traffic-management-shared) to modules/ui/app/scripts/plugins instead of the app folder.

What I want to say: it is working now!

PR #110, opened 2016-10-14
Improve performance of Card View feature

longdogz:
This adds an API that retrieves all the information needed to produce the Card View in one call.  Something that could take 10s or 100s of calls in the old implementation.  The API is also paginated, so that is another performance improvement.  This pull request also includes the changes to the UI to take advantage of this new API.


modules/analytics-objects/src/test/java/com/intuit/wasabi/analyticsobjects/wrapper/ExperimentDetailTest.java
```diff
@@ -0,0 +1,205 @@
...
+
+    Experiment.ID expId = Experiment.ID.newInstance();
+    Experiment.Label expLabel = Experiment.Label.valueOf("TestLabelForExperimentDetail");
+    Experiment.State expState = Experiment.State.RUNNING;
+    Application.Name appName = Application.Name.valueOf("TestApplicationLabelForExperimentDetail");
+    Calendar modTime = Calendar.getInstance();
+    Calendar startTime = Calendar.getInstance();
+    Calendar endTime = Calendar.getInstance();
+
+    {
```
shoeffner: Can this be moved into an `@Before`-method?

modules/analytics/src/main/java/com/intuit/wasabi/analytics/impl/ExperimentDetailsImpl.java
```diff
@@ -0,0 +1,187 @@
...
+     *
+     * @param cassandraRepository repository for the experiment information
+     * @param buckets             access to the bucket information
+     * @param analytics           the analytics module that holds the methods to get bucket details and counts
+     *                            for running experiments
+     */
+    @Inject
+    public ExperimentDetailsImpl(@CassandraRepository ExperimentRepository cassandraRepository,
+                                 Buckets buckets, Analytics analytics) {
+        super();
```
shoeffner: Unnecessary call to `super();`

modules/api/src/main/java/com/intuit/wasabi/api/pagination/filters/impl/ExperimentFilter.java
```diff
@@ -52,7 +52,7 @@ public ExperimentFilter() {
                 Property.creation_time, Property.start_time, Property.end_time, Property.modification_time,
                 Property.date_constraint_start, Property.date_constraint_end);
         super.excludeFromFulltext(Property.application_name_exact, Property.state_exact, Property.date_constraint_start,
-                Property.date_constraint_end);
+                Property.date_constraint_end, Property.favorite);
```
shoeffner: Good find!

modules/api/src/test/java/com/intuit/wasabi/api/pagination/filters/impl/ExperimentDetailFilterTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+        HashMap<String, Boolean> testCases = new HashMap<>();
+
+        testCases.put("experiment_label=ExperimentLabel", true);
+        testCases.put("bucket_label=Bucket2", true);
+        testCases.put("application_name=testApp", false);
+        testCases.put("mod_time=Summer", false);
+        testCases.put("Experiment", true);
+
+        for (Map.Entry<String, Boolean> testCase : testCases.entrySet()) {
```
shoeffner: We should consider changing these (and the tests in the other Filters/Comparators) into Parameterized tests.

AndreaSuckro: I am using this now as a poc in the experimentDetailFilterTest, but didn't want to rewrite the other tests for now. In general I think we should - it is really nicer.

AndreaSuckro:
Thanks :) Now we are just waiting for @evgavry to give her 👍 and then we can merge.

evgavry:
tested it. looks good. Thanks! :shipit:

PR #111, opened 2016-10-15
Feature/rt ingestion json

shoeffner:
Added versioning and proper JSON conversion (no more manual processing) to data ingestion exports.

The resulting JSONs are (examples):

Event (minimal):

``` json
{
    "applicationName": "Application-0", 
    "assignment": {
        "applicationName": null, 
        "bucketEmpty": false, 
        "bucketLabel": null, 
        "cacheable": null, 
        "context": null, 
        "created": null, 
        "experimentID": "3bac275d-96a0-4482-bef8-17906a811fe5", 
        "status": null, 
        "userID": null
    }, 
    "event": {
        "context": {
            "context": "PROD"
        }, 
        "name": "IMPRESSION", 
        "payload": "", 
        "timestamp": 1476551094835, 
        "type": "IMPRESSION"
    }, 
    "experimentLabel": "Experiment-0", 
    "messageType": "EVENT", 
    "timeUUID": "7ea0d3c0-92f9-11e6-b684-ab4174c1be6a", 
    "version": "3.0"
}
```

Event (fully specified):

``` json
{
    "applicationName": "Application-1", 
    "assignment": {
        "applicationName": "Application-1", 
        "bucketEmpty": true, 
        "bucketLabel": "Bucket-1", 
        "cacheable": true, 
        "context": {
            "context": "PREPROD"
        }, 
        "created": 1476550975039, 
        "experimentID": "75b90cdb-6178-47ff-ad25-d83b2013d8f6", 
        "status": "EXISTING_ASSIGNMENT", 
        "userID": "User-1"
    }, 
    "event": {
        "context": {
            "context": "PREPROD"
        }, 
        "name": "CLICKED_BUTTON", 
        "payload": "{\"state\":\"CA\",\"visited\":[\"home\",\"about\"]}", 
        "timestamp": 1476551095040, 
        "type": "BINARY_ACTION"
    }, 
    "experimentLabel": "Experiment-1", 
    "messageType": "EVENT", 
    "timeUUID": "7ea17000-92f9-11e6-b684-ab4174c1be6a", 
    "version": "3.0"
}
```

Assignment (minimal):

``` json
{
    "applicationName": "Application-0", 
    "assignmentStatus": "EXISTING_ASSIGNMENT", 
    "assignmentTimestamp": 1476550866748, 
    "bucketLabel": "Bucket-0", 
    "context": {
        "context": "PROD"
    }, 
    "createAssignment": true, 
    "experimentID": "d8151da4-84b8-4a99-b116-9b5497120264", 
    "experimentLabel": "Experiment-0", 
    "ignoreSamplingPercent": true, 
    "messageType": "ASSIGNMENT", 
    "pageName": "Page", 
    "putAssignment": true, 
    "segmentationProfile": {
        "profile": {}
    }, 
    "timeUUID": "f6a482a0-92f8-11e6-89c6-ab4174c1be6a", 
    "userID": "User 0", 
    "version": "3.0"
}
```

Assignment (fully specified):

``` json
{
    "applicationName": "Application-1", 
    "assignmentStatus": "EXISTING_ASSIGNMENT", 
    "assignmentTimestamp": 1476550866890, 
    "bucketLabel": "Bucket-1", 
    "context": {
        "context": "PROD"
    }, 
    "createAssignment": true, 
    "experimentID": "fe649564-2804-4481-b36d-781ab0dd3e1f", 
    "experimentLabel": "Experiment-1", 
    "ignoreSamplingPercent": true, 
    "messageType": "ASSIGNMENT", 
    "pageName": "Page", 
    "putAssignment": true, 
    "segmentationProfile": {
        "profile": {
            "salary": 30000, 
            "state": "CA", 
            "visited": "home"
        }
    }, 
    "timeUUID": "f6a482a1-92f8-11e6-89c6-ab4174c1be6a", 
    "userID": "User 1", 
    "version": "3.0"
}
```

The segmentationProfile and the context are a little bit overly nested because of how we represent them internally. If you want me to flatten it, we can do it - or defer it to a possible version 4.


modules/analytics-objects/src/main/java/com/intuit/wasabi/analyticsobjects/Event.java
```diff
@@ -72,11 +74,16 @@ public Type getType() {
...
     public void setName(Event.Name value) {
+        if (Objects.isNull(value)) {
+            throw new IllegalArgumentException("Event.Name must not be null.");
+        }
         name = value;
-        if (name.toString().equals(IMPRESSION)) {
-            type = Type.IMPRESSION;
-        } else { //todo: for other event types, will need to check this.value
-            type = Type.BINARY_ACTION;
+        switch (name.toString()) {
```
AndreaSuckro: just wondering why Event.Name never was an enum to begin with...

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -415,9 +415,9 @@ protected Assignment getAssignment(User.ID userID, Application.Name applicationN
 			executors.get(name).execute(new AssignmentEnvelopePayload(userID, context, createAssignment, false,
                     ignoreSamplingPercent, segmentationProfile, assignment != null ? assignment.getStatus() : null,
                     assignment != null ? assignment.getBucketLabel() : null, pageName, applicationName, experimentLabel,
-                    experimentID, currentDate, headers));
-		}        
-        
+                    experimentID, currentDate));
```
AndreaSuckro: unnecessary linebreak?

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -587,9 +587,9 @@ public Assignment putAssignment(User.ID userID, Application.Name applicationName
 			executors.get(name).execute(new AssignmentEnvelopePayload(userID, context, false, true, false, null,
                     Assignment.Status.NEW_ASSIGNMENT, assignment.getBucketLabel(), null, applicationName,
                     experimentLabel, experimentID,
-                    date, null));
-		}        
-        
+                    date));
```
AndreaSuckro: same here

modules/analytics-objects/src/main/java/com/intuit/wasabi/analyticsobjects/Event.java
shoeffner: It is because an action can be "Like button clicked", "Canceled",... i.e. there can be different actions. We just have no analytics for other than binary actions, thats why we only support those two types. At least that's what I assume.

shoeffner:
Thanks for the review @AndreaSuckro, adjusted the code accordingly and went through the `AssignmentsImpl` to get some more of those line breaks streamlined. Also I addressed some FIXMEs in there, mainly removing the `EventLog`.

AndreaSuckro:
:shipit:

mans4singh:
@shoeffner - I think we should also split the code cleaup changes from the required json changes.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -391,46 +394,40 @@ protected Assignment getAssignment(User.ID userID, Application.Name applicationN
                     //todo: change so this doesn't follow the 'read then write' Cassandra anti-pattern
                     assignment = generateAssignment(experiment, userID, context, selectBucket, bucketList, currentDate);
                     assert assignment.getStatus() == Assignment.Status.NEW_ASSIGNMENT :
-                            new StringBuilder("Assignment status should have been NEW_ASSIGNMENT for ")
-                                    .append("userID = \"").append(userID).append("\", experiment = \"")
-                                    .append(experiment).append("\"").toString();
+                            "Assignment status should have been NEW_ASSIGNMENT for " +
```
mans4singh: I think string builder should be used here rather than changing it to string concatenation.

shoeffner: I don't see why we should. The compiler will transform it under the hood anyway: http://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.18.1 , also http://stackoverflow.com/a/1532499 .

shoeffner:
@mans4singh 
> Do we need to have assignments and event together in one json object for events ? Is this a copy and paste issue ?

I don't see what you mean exactly, we have been using assignment information all along to get the user information for an event.

> Why do we need to have context nested under context element ?

Because context is a String inside a Context object.

> Do we need to break down the segmentationProfile into it's subelements ?

We don't, it's just a map of the name "profile" inside an object wrapper.

> Let's discuss these changes.

I can try flattening everything, but this was the simplest way to do it to keep deserialization and serialization reversible.

> I think we should also split the code cleaup changes from the required json changes.

This will take some time.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
mans4singh: Learned something here - thanks.

mans4singh:
> Do we need to have assignments and event together in one json object for events ? Is this a copy and paste issue ?

I think we are adding more attributes of the assignment to the event and which are available in the assignments table.

> Why do we need to have context nested under context element ?

I think we should probably keep it simple and flatten it.

> Do we need to break down the segmentationProfile into it's subelements ?

At the moment, segmentationProfile field maps to one column in the backend so I believe we can keep this value as a blob (json in this case).  

> I think we should also split the code cleaup changes from the required json changes.

The reason I am suggesting this is so that we can roll back changes if required without losing the cleanup changes.  This will also make review/qa easier.

shoeffner:
I close this pull request in favor of future smaller ones with different changes.

PR #123, opened 2016-11-03
Feature/improveTestQuality

AndreaSuckro:
This branch tackles the problem raised by utobi in the context of the  X-Pod code review concerning useless unittests that check for notNull instead of concrete values.


jwtodd:
hey @AndreaSuckro 

nice! i <3 the "make great again" commit message :)

also, have you considered leveraging hamcrest [ https://code.google.com/archive/p/hamcrest/wikis/Tutorial.wiki ] vs straight up junit? i find the combination of junit/hamcrest/mockito to be optimal re test agility/readability, etc.

modules/api/src/test/java/com/intuit/wasabi/api/AssignmentsResourceTest.java
```diff
@@ -63,7 +66,9 @@
     @Mock
     private ExperimentBatch experimentBatch;
     @Mock
-    private @javax.ws.rs.core.Context HttpHeaders headers;
+    private
+    @javax.ws.rs.core.Context
```
jwtodd: we really want to spread this declaration across multiple lines?

modules/api/src/test/java/com/intuit/wasabi/api/AssignmentsResourceTest.java
```diff
@@ -203,6 +208,6 @@ public void postBatchAssignmentForPage() throws Exception {
 
     @Test
     public void getAssignmentsQueueLength() throws Exception {
-        assertNotNull(resource.getAssignmentsQueueLength());
+        assertTrue(resource.getAssignmentsQueueLength().getStatus() == HttpStatus.SC_OK);
```
jwtodd: a hamcrest equivalent would be, psuedocode:

  assertThat(resource.getAssignmentQueueLength(), is(SC_OK))

what say you?

jwtodd: @AndreaSuckro i can help with this patch if you'd like, add some hamcrest equivalents?

AndreaSuckro: I will look into hamcrest a bit more and send an update (?) Thank you very much for the feedback! ❤️

AndreaSuckro:
Hey @jwtodd : I updated some of the test classes to now rely completely on hamcrest. I think it would be nice to have a complete alignment of all the unit tests, but I can't provide that at the moment really. 

While writing the code some questions came to my mind: does it make sense to have a mixed style of junit and hamcrest asserts? Because in the beginning it wasn't really clear to me why for example
`assertNotNull(val)` is worse to read then `assertThat(val, notNullValue())`. On the other hand: using only hamcrest would allow for more consistency and less dependencies in the code. What do you think?

PR #124, opened 2016-11-04
mod: jacoco exclude (modules/functional-tests)

jwtodd:
goal: filter modules/functiona-tests from jacoco analytics

results: 64.26% unit test coverage, up from 35%

cc: @iizrailevsky @AndreaSuckro @shoeffner


iizrailevsky:
Awesome @jwtodd! We're back in business. 👍 :shipit:

PR #128, opened 2016-11-11
Feature/datastax-missingcases(getExpByApp,getExpByPage)

sudheerkethireddy:


modules/functional-test/src/main/java/com/intuit/wasabi/tests/library/TestBase.java
```diff
@@ -3075,6 +3126,60 @@ public Response putApplicationPriorities(Application application, List<Experimen
...
+    /**
+     * 
+     * Sends a GET request to receive a list of experiments for an application
+     * The response must contain HTTP {@link HttpStatus#SC_OK}
+     * 
+     * @param application the application for which the experiments are
+     * @param expectedStatus the expected HTTP status code
+     * @return a list of experiments
+     */
+    public List<Experiment> getExperimentsByApplication(Application application, int expectedStatus)
```
mans4singh: Are we using expectedStatus in this call ?

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/IntegrationExperiment.java
```diff
@@ -681,6 +685,33 @@ public void t_createTestPersonalizationExperiment() {
...
+    }
+     
+    /**
+     * This test case covers a scenario where we
+     * try to get list of experiments of  
+     * non-existent application and non-existent page
+     * the name of the app I am using is junkapp
+     * the name of the page I am using is junkpage
+     */
+    @Test
```
mans4singh: Do we have positive test cases ?

mans4singh:
Looks good to me.

PR #131, opened 2016-11-16
Fix bug that if you filter in the Experiments list or Card View and t…

longdogz:
…hen click on a Running experiment from the results, when you come back, the filter is not applied.  Also fixed that if you cleared the search box with a backspace, that didn’t clear the filtering.


evgavry:
tested it. Looks good. :shipit: thanks!

PR #133, opened 2016-11-17
Feature/card view filtering

longdogz:

PR #135, opened 2016-11-19
experiment_label consistently for sorting and filtering

shoeffner:
while fixing broken integration tests introduced with #133, I made the behaviour for pagination and filtering more consistent by always using experiment_label instead of experiment_name.


iizrailevsky:
Looks good @shoeffner - thanks for fixing this! 👍  :shipit:

PR #136, opened 2016-11-21
Disabling auto-gen attributions plugin for now to minimize build time

iizrailevsky:


felixgao:
other than that :shipit:

PR #137, opened 2016-11-21
URL encoding the parameter for free text search, as providing an equa…

longdogz:
…ls sign caused a 500.  Also, added checks for empty dates on the advanced search and showing an error in that case.


evgavry:
tested it. looks good. :shipit:

PR #143, opened 2016-11-30
added description field to bucketdetail and experimentdetail

AndreaSuckro:
Adding the description value to the card view api.


evgavry:
tested :shipit: thanks!

PR #144, opened 2016-12-01
Added get assignment test cases for empty bucket

mans4singh:


felixgao:
not sure if we want to this behavior for the open source version as well. But it looks good if we are all ok to have this behavior only in intuit specific code

mans4singh:
The functional tests are in open source branch that is what this pull request includes.

mans4singh:
Just a clarification - this functionality is required in open source and wasabi-intuit.  In this checkin there are unit tests for the empty bucket reassignment use case.

PR #145, opened 2016-12-21
Feature/json export

shoeffner:
Changes the json export structure to be flat and to only rely on well tested frameworks (Jackson in our case). It also introduces a version number to track changes and adjust downstream mechanisms accordingly. Since we already internally iterated a number of versions, events start at version 3.0 and assignments at version 4.0.

Examples are given below.

**Events**

    {  
        "applicationName":"a1",
        "experimentLabel":"l1",
        "event":{  
            "name":"CLICKED_BUTTON",
            "type":"BINARY_ACTION",
            "payload":"SomePayload{}\\\\!@#\"\"withEscaping",
            "timestamp":1482316692684
        },
        "timeUUID":"92c50de0-c769-11e6-9dbf-8118b1965c72",
        "value":null,
        "context":"PROD",
        "messageType":"EVENT",
        "userID":"TestUser",
        "experimentID":"df910303-ec84-4f95-9608-2c84b8daad2c",
        "bucketLabel":"blue",
        "version":"3.0"
    }

**Assignments**

```
{  
    "userID":"testUser",
    "context":"testContext",
    "createAssignment":true,
    "putAssignment":true,
    "ignoreSamplingPercent":true,
    "segmentationProfile":"{}",
    "assignmentStatus":"EXISTING_ASSIGNMENT",
    "bucketLabel":"testLabel",
    "pageName":"testPage",
    "applicationName":"testAppName",
    "experimentLabel":"testExperimentLabel",
    "experimentID":"81f8c201-cedd-4208-b248-b0d3ef8d20d2",
    "timeUUID":"d9a4ab80-c769-11e6-809e-8118b1965c72",
    "assignmentTimestamp":1482316811523,
    "version":"4.0",
    "messageType":"ASSIGNMENT"
}
```

**Assignments 2**

```
{  
    "userID":"testUser",
    "context":"testContext",
    "createAssignment":true,
    "putAssignment":false,
    "ignoreSamplingPercent":true,
    "segmentationProfile":"{\"visited\":\"home\",\"state\":\"CA\",\"salary\":30000}",
    "assignmentStatus":"EXISTING_ASSIGNMENT",
    "bucketLabel":"testLabel",
    "pageName":"testPage",
    "applicationName":"testAppName",
    "experimentLabel":"testExperimentLabel",
    "experimentID":"81f8c201-cedd-4208-b248-b0d3ef8d20d2",
    "timeUUID":"d9e006d0-c769-11e6-809e-8118b1965c72",
    "assignmentTimestamp":1482316811523,
    "version":"4.0",
    "messageType":"ASSIGNMENT"
}
```


AndreaSuckro:
I have pushed a suggestion for resolving the circular reference. I know it's not directly pretty, but here is why I think it is ok in this case: the only reason why the dependency existed in the first place was the reference to User.ID, but this is not a useful class, since it is just one of these wrappers around a String which we want to tackle shortly anyways. This is why I thought it is okay to replace the direct reference and just use plain Strings in the exceptions itself.

AndreaSuckro:
@mooxter we have to check how to integrate all the new code into this branch and what the implications of the altered structure are to the clients. Otherwise these changes make a lot of sense.

shoeffner:
Closing as stale.

PR #147, opened 2017-01-09
initial removal of user-assignment and user-assignment-index accessors

felixgao:


modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -219,7 +219,7 @@ public Assignment getSingleAssignment(User.ID userID, Application.Name applicati
                     Assignment.Status.EXPERIMENT_EXPIRED);
         }
 
-        Assignment assignment = assignmentsRepository.getAssignment(experimentID, userID, context);
+        Assignment assignment = assignmentsRepository.getAssignment(experimentID, applicationName, userID, context);
```
jcwuzoegiver: just to test my understanding. applicationName is added because it's need to query experiment_user_index table. Am I correct?

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/AssignmentsRepository.java
```diff
@@ -107,6 +107,7 @@ void deleteAssignment(Experiment experiment, User.ID userID, Context context, Ap
      * @param date       Date of assignment
      * @return Assignment
      */
+    @Deprecated
```
jcwuzoegiver: is the idea to deprecate this method so it can be removed later?

pom.xml
```diff
@@ -77,8 +77,6 @@ http://www.w3.org/2001/XMLSchema-instance">
         <authentication.class.name>com.intuit.wasabi.authentication.impl.DefaultAuthentication
         </authentication.class.name>
 
-        <assign.user.to.old>true</assign.user.to.old>
-        <assign.user.to.new>false</assign.user.to.new>
```
jcwuzoegiver: with the removal of these 2 props, it will be silently switched to the new table once we deploy this new code. since we are not writing to the old table anymore, there will be no way to roll back to previous code if things go wrong.

PR #148, opened 2017-01-09
Externalize installing modules so it can be overridden by child class.

jcwuzoegiver:
@felixgao @mans4singh this is part of the dedupe changes that will is needed for wasabi-intuit. [cc] @iizrailevsky 
@AndreaSuckro this change will fix the extra bindings introduced in ApiModule for the pagination changes in develop.

### DIFF
--- a/modules/api/src/main/java/com/intuit/wasabi/api/ApiModule.java
+++ b/modules/api/src/main/java/com/intuit/wasabi/api/ApiModule.java
@@ -55,22 +55,8 @@ public class ApiModule extends AbstractModule {
 
     @Override
     protected void configure() {
-        LOGGER.debug("installing module: {}", ApiModule.class.getSimpleName());
-        //these modules are either free of other dependencies or they are required by later modules
-        install(new com.intuit.autumn.api.ApiModule());
-        install(new UserDirectoryModule());
-        install(new DatabaseExperimentRepositoryModule());
-        install(new DatabaseModule());
-        install(new JacksonModule());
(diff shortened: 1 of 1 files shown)


modules/api/src/main/java/com/intuit/wasabi/api/ApiModule.java
```diff
@@ -101,6 +87,26 @@ protected void configure() {
 //            e.printStackTrace();
 //        }
 
+    }
+    
+    protected void installModules() {
+        LOGGER.debug("installing module: {}", ApiModule.class.getSimpleName());
+
```
mans4singh: We have multiple ApiModule classes - so perhaps we can use the fully qualified name here.

jcwuzoegiver: Will change it to ApiModule.class..getCanonicalName()

modules/api/src/main/java/com/intuit/wasabi/api/ApiModule.java
```diff
@@ -101,6 +87,38 @@ protected void configure() {
 //            e.printStackTrace();
 //        }
 
-        LOGGER.debug("installed module: {}", ApiModule.class.getSimpleName());
+    }
+    
+    protected void installUserModule() {
```
felixgao: I am assuming you are doing this because you want to overwrite the behavior when intuit specific class going to install different modules. Isn't that already controlled by the properties file?

PR #149, opened 2017-01-10
Code dedupe part 1 (api module)

jcwuzoegiver:
Refactored ApiModule so common module installation will not be repeated in child class and some modules installation can be overridden.
@mans4singh @felixgao


felixgao:
:shipit:

mans4singh:
:shipit:

PR #158, opened 2017-01-18
adding missing licence information for new classes

AndreaSuckro:


modules/repository-datastax/integration-test/test/java/com/intuit/wasabi/repository/cassandra/impl/authorization/AuthorizationRepositorySetup.java
```diff
@@ -22,34 +37,40 @@
...
 
     protected final Set<String> apps = new HashSet<>();
     protected final Map<String, Set<String>> appUser = new HashMap<>();
 
-    public void setupDb(String appName, String username, String role){
+    public void setupDb(String appName, String username, String role) {
         UserInfo userInfo = userDirectory.lookupUser(UserInfo.Username.valueOf(username));
-        logger.info(appName+" "+" "+username+" "+role + " "+userInfo);
-        if(!"*".equals(appName)){
+        logger.info(appName + " " + " " + username + " " + role + " " + userInfo);
```
mans4singh: Can we avoid string concatination ?

modules/repository-datastax/integration-test/test/java/com/intuit/wasabi/repository/cassandra/impl/authorization/AuthorizationRepositorySetup.java
```diff
@@ -22,34 +37,40 @@
...
+        if (!"*".equals(appName)) {
```
mans4singh: Extract "*" as constant.

modules/repository-datastax/src/test/java/com/intuit/wasabi/repository/cassandra/accessor/ExperimentAccessorITest.java
```diff
@@ -16,82 +35,84 @@
...
     static Date date1 = new Date();
     static Date date2 = new Date();
-    
+
     @BeforeClass
-    public static void setup(){
-    	IntegrationTestBase.setup();
-    	if (accessor != null) return;
+    public static void setup() {
+        IntegrationTestBase.setup();
```
mans4singh: Let's keep formatting changes separate from code/related changes.

jcwuzoegiver:
@AndreaSuckro it will be good not to change the format of the code if you're just adding license headers. It's harder to review the code. Formatting code can be considered in separate improvement. [cc] @bargenilesh @mans4singh

AndreaSuckro:
Hm you are right - there is actually a PR (https://github.com/intuit/wasabi/pull/159) for that aswell ;) Would be nice if you guys could take a look there aswell, if everyone uses the same formatter and we can agree on a format that makes sense all troubles will be (hopefully) gone. My editor is set in a way that it automatically formats stuff on save, but I can disable it and add the headers again.

modules/repository-datastax/integration-test/test/java/com/intuit/wasabi/repository/cassandra/impl/authorization/AuthorizationRepositorySetup.java
AndreaSuckro: @mans4singh those are all good comments we should keep in mind, but let's do only the headers here. Will update the PR shortly

PR #159, opened 2017-01-18
adding codestyle formatters for java

AndreaSuckro:
To speed up our peer review process and to make the code more consistent it would be nice to have one codestyle that the developers can use in their IDE. This branch provides a formatter for both IntelliJ and Eclipse and ran it on our current code base. This is currently just limited to the java classes, but could of course also be applied to the UI code with an own set of rules.


iizrailevsky:
Looks good :shipit:

bargenilesh:
@AndreaSuckro Looks good 👍, provided unit & functional tests and passing for both wasabi & wasabi-intuit :shipit:

jcwuzoegiver:
@AndreaSuckro Looks good :shipit:. Now we need to enforce developers to use the codestyle formatters in */codestyle/*.xml, right? Do you have instructions to do so? I imagine you will need to do the same thing for wasabi-intuit.

PR #160, opened 2017-01-18
dedupe assignment module

jcwuzoegiver:
1. Consolidated 2 generateAssignment() methods into one.
2. Eliminated redundant code and unused
mergePersonalizationResponseWithSegmentation() method
3. Updated getBucketList() method so it will call AssignmentDecorator to
retrieve recommended bucketList.

@mans4singh @bargenilesh

### DIFF
--- a/modules/assignment/src/main/java/com/intuit/wasabi/assignment/AssignmentDecorator.java
+++ b/modules/assignment/src/main/java/com/intuit/wasabi/assignment/AssignmentDecorator.java
@@ -15,23 +15,38 @@
  *******************************************************************************/
 package com.intuit.wasabi.assignment;
 
+import java.io.IOException;
+
+import com.intuit.wasabi.assignmentobjects.SegmentationProfile;
+import com.intuit.wasabi.assignmentobjects.User;
+import com.intuit.wasabi.exceptions.BucketDistributionNotFetchableException;
+import com.intuit.wasabi.experimentobjects.BucketList;
 import com.intuit.wasabi.experimentobjects.Experiment;
 
-import java.io.UnsupportedEncodingException;
-import java.net.URI;
-
 /**
- * Decorator for creating uri from experiment
+ * Interface that provides method(s) to modify/enhance the default behavior of retrieving bucket list when
+ * retrieving assignment(s).
  */
 public interface AssignmentDecorator {
 
-
     /**
-     * Given an experiment, derive the materialized URI for calling other resources on this experiment
+     * Gets the bucket list through an external source (AssignmentDecorator). AssignmentDecorator can
+     * be a personalization engine that will execute a model to return a recommended bucket list. If the call
+     * is a successful, it merges the personalization parameters with segmentation profile
      *
-     * @param experiment {@link com.intuit.wasabi.experimentobjects.Experiment} current experiment
-     * @return URI contains the materialized URI for experiment
-     * @throws UnsupportedEncodingException
-     */
-    URI materializeUri(Experiment experiment) throws UnsupportedEncodingException;
+     * @param experiment
+     *            Experiment , it provides the modelName and modelVersion that needs to be used
+     * @param userID
+     *            UserId
+     * @param segmentationProfile
+     *            (provides the segmentationProfile to personalize the assignment)
+     * @return BucketList, which contains the buckets and the respective
+     *         allocation percentages for a userID
+     * @throws IOException
+     *             IO exception
+     * @throws com.intuit.wasabi.exceptions.BucketDistributionNotFetchableException
+     *             not able to fetch bucket distribution
+     */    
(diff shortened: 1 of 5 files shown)


modules/assignment/src/test/java/com/intuit/wasabi/assignment/impl/DefaultAssignmentDecoratorTest.java
```diff
@@ -37,11 +40,12 @@
         assertThat(defaultAssignmentDecorator, is(not(nullValue())));
     }
 
-    @Test public void materializeUriTest() throws UnsupportedEncodingException {
+    @Test public void getBucketListTest() throws UnsupportedEncodingException {
```
mans4singh: Do need test for positive case (when bucket list returned is not null) ?

modules/assignment/src/main/java/com/intuit/wasabi/assignment/AssignmentDecorator.java
```diff
@@ -15,23 +15,38 @@
...
-import java.net.URI;
+import com.intuit.wasabi.assignmentobjects.SegmentationProfile;
+import com.intuit.wasabi.assignmentobjects.User;
+import com.intuit.wasabi.exceptions.BucketDistributionNotFetchableException;
+import com.intuit.wasabi.experimentobjects.BucketList;
+import com.intuit.wasabi.experimentobjects.Experiment;
 
 /**
- * Decorator for creating uri from experiment
+ * Interface that provides method(s) to modify/enhance the default behavior of retrieving bucket list when
```
mans4singh: Should we name the interface BucketSelector - since it is just getting a bucket list ?

jcwuzoegiver: @mans4singh probably not. Today this interface AssignmentDecorator can "decorate" bucket list. Tomorrow, we can add another method to "decorate" the assignment. Let's keep the same name for now. agreed?

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -963,24 +939,46 @@ private Assignment generateAssignment(Experiment experiment, User.ID userID, Con
...
      */
-    protected BucketList getBucketList(Experiment experiment, Boolean skipBucketRetrieval) {
-        BucketList buckets = null;
-        if (!skipBucketRetrieval) {
-            buckets = repository.getBuckets(experiment.getID());
+    protected BucketList getBucketList(Experiment experiment, User.ID userID, SegmentationProfile segmentationProfile, Boolean skipBucketRetrieval) {
+        
+        BucketList bucketList = null;
+        Boolean isPersonalizationEnabled = experiment.getIsPersonalizationEnabled();
+        if (isPersonalizationEnabled && !Objects.isNull(assignmentDecorator)) {
```
bargenilesh: Very minor: No need of this condition "!Objects.isNull(assignmentDecorator)" as assignmentDecorator should not be NULL in any case...

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -963,24 +939,46 @@ private Assignment generateAssignment(Experiment experiment, User.ID userID, Con
...
+            try {
+                bucketList = assignmentDecorator.getBucketList(experiment, userID, segmentationProfile);
+            } catch (BucketDistributionNotFetchableException | IOException e) {
+                LOGGER.error(new StringBuilder("Error obtaining the Bucket List from ").append(assignmentDecorator.getClass().getSimpleName())
```
bargenilesh: Can we add 'e' exception variable as second parameter to the LOGGER.error() method, that will print complete stack trace which will be helpful for investigation...

modules/assignment/src/test/java/com/intuit/wasabi/assignment/impl/DefaultAssignmentDecoratorTest.java
jcwuzoegiver: @mans4singh DefaultAssignmentDecorator's getBucketList() always return null;

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
jcwuzoegiver: @bargenilesh going to eliminate IOException altogether from AssignmentDecorator.getBucketList() interface. This method should not throw IOException. It should be abstract out from the actual implementation of this method by sub class.

mans4singh:
:shipit:

PR #161, opened 2017-01-18
adding licence without code formatting

AndreaSuckro:
This time without cluttered codeformat changes (although we should really straighten this I think - some of the files are even indented with tabs etc.). But let https://github.com/intuit/wasabi/pull/159 worry about this.


jcwuzoegiver:
@AndreaSuckro  looks good w/o code formatting. :shipit:

PR #162, opened 2017-01-18
Feature/remove unused tables - Drop user_assignment_look_up table and user_experiment_index table

mans4singh:


modules/repository-datastax/db/mutation/V030__Drop_User_Assignment_Look_Up_table.cql
```diff
@@ -0,0 +1 @@
+drop table user_assignment_look_up;
```
bargenilesh: Would it be better to use "DROP TABLE IF EXISTS keyspace_name.table_name" syntax?

modules/repository-datastax/db/mutation/V031__Drop_User_Experiment_Index_table.cql
```diff
@@ -0,0 +1 @@
+drop table user_experiment_index;
```
bargenilesh: Would it be better to use "DROP TABLE IF EXISTS keyspace_name.table_name" syntax?

bargenilesh: Also, is there anyway to set consistency to ALL while executing these commands or is it something already taken care in the migration-script code (caller of these .cql files).

PR #164, opened 2017-01-19
Eliminate unnecessary multiple getExperiment calls

jcwuzoegiver:
This change is to eliminate the unnecessary multiple calls to get experiment from repository (Cassandra) and therefore it will improve overall performance of single assignment call, batch assignment call and other call that requires retrieving of bucket list. For example, the following methods retrieve experiment from repository more than 1 time in the current code base.

AssignmentImpl->getSingleAssignment
AssignmentImpl->doBatchAssignments
ExperimentResource->putBucket()
BucketsImpl->createBucket()
..
..

This is not an attempt to use Datastax async call for single assignment to improve performance. I manually traced all call stacks to getBuckets(Experiment.ID experimentId). Only 1 caller requires to check/retrieve experiment again. This method is now changed to getBuckets(Experiment.ID experimentId, boolean checkExperiment).

### DIFF
--- a/modules/analytics/src/main/java/com/intuit/wasabi/analytics/impl/AnalyticsImpl.java
+++ b/modules/analytics/src/main/java/com/intuit/wasabi/analytics/impl/AnalyticsImpl.java
@@ -528,12 +528,12 @@ public class AnalyticsImpl implements Analytics {
 
         // Uses counters
         Experiment experiment = cassandraRepository.getExperiment(experimentID);
+        if (Objects.isNull(experiment)) {
+            throw new ExperimentNotFoundException(experimentID);
+        }            
         if (release_date != null && release_date.before(experiment.getCreationTime())) {
             return assignmentRepository.getBucketAssignmentCount(experiment);
         } else {
-            if (Objects.isNull(experiment)) {
-                throw new ExperimentNotFoundException(experimentID);
-            }            
             return cassandraRepository.getAssignmentCounts(experimentID, context);
         }
     }
--- a/modules/analytics/src/main/java/com/intuit/wasabi/analytics/impl/AnalyticsImpl.java
+++ b/modules/analytics/src/main/java/com/intuit/wasabi/analytics/impl/AnalyticsImpl.java
@@ -531,6 +531,9 @@ public class AnalyticsImpl implements Analytics {
         if (release_date != null && release_date.before(experiment.getCreationTime())) {
             return assignmentRepository.getBucketAssignmentCount(experiment);
         } else {
+            if (Objects.isNull(experiment)) {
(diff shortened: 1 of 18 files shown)


jcwuzoegiver:
Thanks @AndreaSuckro. @iizrailevsky @bargenilesh can you pls review too?

modules/analytics/src/main/java/com/intuit/wasabi/analytics/impl/AnalyticsImpl.java
```diff
@@ -531,6 +531,9 @@ public AssignmentCounts getAssignmentCounts(Experiment.ID experimentID, Context
         if (release_date != null && release_date.before(experiment.getCreationTime())) {
             return assignmentRepository.getBucketAssignmentCount(experiment);
         } else {
+            if (Objects.isNull(experiment)) {
```
mans4singh: Perhaps we can check for experiment to be null before the if statement.

modules/analytics/src/main/java/com/intuit/wasabi/analytics/impl/ExperimentDetailsImpl.java
```diff
@@ -89,7 +89,7 @@ public ExperimentDetailsImpl(@CassandraRepository ExperimentRepository cassandra
      * @return the same ExperimentDetail object but with additional information
      */
     private ExperimentDetail getBucketData(ExperimentDetail exp) {
-        List<Bucket> buckList = buckets.getBuckets(exp.getId()).getBuckets();
+        List<Bucket> buckList = buckets.getBuckets(exp.getId(), false).getBuckets();
```
mans4singh: I think changing the name of the method to getBucketsWith/WithoutExp check is more comprehensible than true/false argument.

modules/analytics/src/main/java/com/intuit/wasabi/analytics/impl/AnalyticsImpl.java
jcwuzoegiver: @mans4singh Make sense. Will check if experiment is null before if statement.

modules/analytics/src/main/java/com/intuit/wasabi/analytics/impl/ExperimentDetailsImpl.java
jcwuzoegiver: @mans4singh it's internal interface. is it really necessary? having an additional method in Buckets Interface will affect both CassandraExperimentRepository and DatabaseExperimentRepository (MySQL).

mans4singh:
Sounds good :shipit:

PR #165, opened 2017-01-20
added QE-automation for userRole Issue

sudheerkethireddy:


modules/functional-test/src/main/java/com/intuit/wasabi/tests/library/TestBase.java
```diff
@@ -89,4297 +91,4493 @@
 @Listeners({RetryListener.class})
 public class TestBase extends ServiceTestBase {
 
-    private static final Logger LOGGER = getLogger(TestBase.class);
```
jcwuzoegiver: @sudheerkethireddy it looks like you formatted the code so it created a big DIFF. Can you not format the code if there is no change in the code?

modules/functional-test/src/main/resources/config.properties
```diff
@@ -39,6 +39,7 @@ api-server-protocol=http
 api-version-string=v1
 user-name=admin
 password=admin
+test-user=wasabi_reader@example.com
```
jcwuzoegiver: @sudheerkethireddy is it possible to reuse the existing user for this test case? this will avoid adding this test-user to the config.properties of <company-specific>-functional-test module.

sudheerkethireddy: @jcwuzoegiver the above is a default user we need to have, because admin is the superuser and he should be able to assign users to the applications. for that he need one non-super admin user and hence i used wasabi_reader@example.com

modules/functional-test/src/main/java/com/intuit/wasabi/tests/library/TestBase.java
sudheerkethireddy: sure @jcwuzoegiver (y)

modules/functional-test/src/main/java/com/intuit/wasabi/tests/library/TestBase.java
```diff
@@ -4336,7 +4338,172 @@ public Response getLogout(AccessToken token, int expectedStatus, APIServerConnec
...
+	  * @param user the user whose list of applications we are interested
+	  * @param expectedStatus  the expected HTTP status code 
+	  * @param apiServerConnector the server connector to use
+	  * @return List of application names
+	  */
+    public List<String>  getUserApplications(APIUser user, int expectedStatus, APIServerConnector apiServerConnector) {
+    	apiServerConnector.setUserNameAndPassword(user.userId, user.password);
+		response = apiServerConnector.doGet("authorization/applications");
+		assertReturnCode(response,expectedStatus);
+		List<Map<String, Object>> jsonStrings = response.jsonPath().get("roleList");
```
mans4singh: Extract string constants (roleList)

modules/functional-test/src/main/java/com/intuit/wasabi/tests/library/TestBase.java
```diff
@@ -4336,7 +4338,172 @@ public Response getLogout(AccessToken token, int expectedStatus, APIServerConnec
...
+	  * @param experiment         the experiment to POST
+	  * @param expectedStatus     the expected HTTP status code
+	  * @param apiServerConnector the server connector to use
+	  * @return List of assignment status
+	  */
+    public List<AssignmentStatus>  postUserRolePermission(List<UserRole> roles, int expectedStatus, APIServerConnector apiServerConnector) {
+    	//building the payload
+    	StringBuilder stringBuilder = new StringBuilder();
+		stringBuilder.append("{");
+		stringBuilder.append("\"roleList\":");
```
mans4singh: Using string concatination is not safe.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/UserPermissionsTest.java
```diff
@@ -0,0 +1,164 @@
...
+	String testUserPassword = null;
+	String testUserID = null;
+	List<Application> applicationList = null;
+
+	@BeforeClass()
+	public void initializeTest()
+	{
+		testUser = appProperties.getProperty("test-user", Constants.DEFAULT_TEST_USER);
+		//lets create applications first
+		Application application1 = ApplicationFactory.createApplication().setName("testApplication_1");
```
mans4singh: We can use a loop here to create experiments.

mans4singh: Or use a helper method to create experiments by arguments

sudheerkethireddy: Sure @mans4singh  thanks for the feedback. will add those changes

modules/functional-test/src/main/java/com/intuit/wasabi/tests/library/TestBase.java
sudheerkethireddy: yes mansi know its ugly but i dont have any other way like Gson where i can construct the payload.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/library/TestBase.java
sudheerkethireddy: sure will do

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/UserPermissionsTest.java
```diff
@@ -0,0 +1,142 @@
...
+		
+		//lets pause the experiments
+		for(int i = 1 ; i <= NUMBER_OF_APPLICATIONS; i++) {
+			Experiment experiment = experimentList.get(i-1);
+			experiment.state = Constants.EXPERIMENT_STATE_PAUSED;
+			putExperiment(experiment);
+		}
+		
+		//once paused, lets terminate the experiments
+		for(int i = 1 ; i <= NUMBER_OF_APPLICATIONS; i++) {
```
mans4singh: Can we put this in the first loop ?  Otherwise looks good.

mans4singh:
:shipit:

evgavry:
:shipit: looks good. Thanks for fixing indentation.

modules/functional-test/src/main/resources/config.properties
jcwuzoegiver: @sudheerkethireddy will there be a test case in the future using wasabi_writer? It will be good to name test-user to test-wasabi-reader. It will be consistent to the userDirectory.properties.

sudheerkethireddy: sure @jcwuzoegiver i will incorporate the change. thanks for the very good suggestion.

PR #166, opened 2017-01-22
fixing the cardview for production

AndreaSuckro:
The basic problem was also an error in the way the experiments where retrieved, since the experimentpojo stored the state only as a string and not as an enum the check for the deleted state could never succeed.


iizrailevsky:
Makes sense @AndreaSuckro, looks good 👍 :shipit:

longdogz:
Looks fine to me.  :shipit:

PR #167, opened 2017-01-23
Fix/cardview

AndreaSuckro:
Corrected the wrong conditional in `setState()` and added a unit test for this case.


jcwuzoegiver:
@AndreaSuckro looks good after updating the logic with new test cases 👍 . :shipit:

iizrailevsky:
👍 :shipit:

PR #170, opened 2017-02-02
Short term fix for memory exhaustion issue due to growing ingestion queue

jcwuzoegiver:
This enhancement is to:
1) Fix memory/resource exhaustion issue due to Ingestion Executor's
queue keep growing when active threads are blocked.
2) Added new admin REST API to flush and persist active & pending
messages in the queue and another API to return detailed of the queue
status.

### DIFF
--- a/modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
+++ b/modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
@@ -23,6 +23,9 @@ import com.intuit.wasabi.assignment.Assignments;
 import com.intuit.wasabi.assignmentobjects.Assignment;
 import com.intuit.wasabi.assignmentobjects.SegmentationProfile;
 import com.intuit.wasabi.assignmentobjects.User;
+import com.intuit.wasabi.authenticationobjects.UserInfo;
+import com.intuit.wasabi.authenticationobjects.UserInfo.Username;
+import com.intuit.wasabi.authorization.Authorization;
 import com.intuit.wasabi.exceptions.AssignmentNotFoundException;
 import com.intuit.wasabi.experimentobjects.*;
 import com.intuit.wasabi.experimentobjects.Bucket.Label;
@@ -43,7 +46,9 @@ import java.util.Map;
 import static com.google.common.collect.Maps.newHashMap;
 import static com.intuit.wasabi.api.APISwaggerResource.*;
 import static com.intuit.wasabi.assignmentobjects.Assignment.Status.EXPERIMENT_EXPIRED;
+import static com.intuit.wasabi.authorizationobjects.Permission.CREATE;
 import static java.lang.Boolean.FALSE;
+import static javax.ws.rs.core.HttpHeaders.AUTHORIZATION;
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 import static org.slf4j.LoggerFactory.getLogger;
 
@@ -59,11 +64,13 @@ public class AssignmentsResource {
     private static final Logger LOGGER = getLogger(AssignmentsResource.class);
     private final HttpHeader httpHeader;
     private final Assignments assignments;
+    private Authorization authorization;
 
     @Inject
-    AssignmentsResource(final Assignments assignments, final HttpHeader httpHeader) {
+    AssignmentsResource(final Assignments assignments, final HttpHeader httpHeader, final Authorization authorization) {
         this.assignments = assignments;
         this.httpHeader = httpHeader;
+        this.authorization = authorization;
     }
 
     /**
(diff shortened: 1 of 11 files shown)


jcwuzoegiver:
@mans4singh @bargenilesh [cc] @iizrailevsky please review.

modules/api/src/test/java/com/intuit/wasabi/api/AssignmentsResourceTest.java
```diff
@@ -217,4 +232,29 @@ public void postBatchAssignmentForPage() throws Exception {
     public void getAssignmentsQueueLength() throws Exception {
         assertThat(resource.getAssignmentsQueueLength().getStatus(), is(HttpStatus.SC_OK));
     }
+    
+    @Test
+    public void getAssignmentsQueueDetails() throws Exception {
+        assertThat(resource.getAssignmentsQueueDetails().getStatus(), is(HttpStatus.SC_OK));
```
mans4singh: Do we need to check the body content ?

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
```diff
@@ -533,6 +540,34 @@ public Response getAssignmentsQueueLength() {
...
+     */
+    @POST
+    @Path("flushMessages")
+    @Produces(APPLICATION_JSON)
+    public Response flushMessages(
+            @HeaderParam(AUTHORIZATION) @ApiParam(value = EXAMPLE_AUTHORIZATION_HEADER, required = true) final String authorizationHeader) {
+        Username userName = authorization.getUser(authorizationHeader);
+        authorization.checkSuperAdmin(userName);
+        assignments.flushMessages();
+        return httpHeader.headers().build();
```
mans4singh: Should the return status be no content ?

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -1095,7 +1104,25 @@ public boolean doSegmentTest(Application.Name applicationName, Experiment.Label
         return queueLengthMap;
     }
 
-
+    @Override
+    public Map<String, Object> queuesDetails() {
+        Map<String, Object> queueLengthMap = new HashMap<String, Object>();
```
mans4singh: Please add logging.

modules/assignment/src/test/java/com/intuit/wasabi/assignment/impl/AssignmentsImplTest.java
```diff
@@ -39,7 +39,6 @@
 import com.intuit.wasabi.repository.MutexRepository;
 import com.intuit.wasabi.repository.cassandra.impl.ExperimentRuleCacheUpdateEnvelope;
 import com.netflix.astyanax.connectionpool.exceptions.ConnectionException;
```
mans4singh: We should remove astyanax exceptions

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
jcwuzoegiver: @mans4singh why do you want logging?

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
jcwuzoegiver: @mans4singh Good catch! Will do.

modules/assignment/src/test/java/com/intuit/wasabi/assignment/impl/AssignmentsImplTest.java
jcwuzoegiver: @mans4singh Will do!

modules/api/src/test/java/com/intuit/wasabi/api/AssignmentsResourceTest.java
jcwuzoegiver: @mans4singh Will check body content as well.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
mans4singh: This was a general comment - it makes debugging easier.

jcwuzoegiver: Ok but I don't think it is necessary because the code is very simple. It just return a Map.

mans4singh:
minor comments about logging - otherwise looks good - :shipit:

PR #171, opened 2017-02-08
Add Access-Control-Max-Age and fix duplicate CORS

jcwuzoegiver:
This request is to do the followings:

1) Add Access-Control-Max-Age response header so client can cache preflight request.
2) Fix duplicate CORS headers when call fails for OPTIONS request.

### DIFF
--- a/modules/api/src/main/java/com/intuit/wasabi/api/HttpHeader.java
+++ b/modules/api/src/main/java/com/intuit/wasabi/api/HttpHeader.java
@@ -30,12 +30,13 @@ import static javax.ws.rs.core.Response.status;
 public class HttpHeader {
 
     private final String applicationName;
+    private final String deltaSeconds;
     private final CacheControl cacheControl;
 
     @Inject
-    public HttpHeader(final @Named("application.id") String applicationName) {
+    public HttpHeader(final @Named("application.id") String applicationName, final @Named("access.control.max.age.delta.seconds") String deltaSeconds) {
         this.applicationName = applicationName;
-
+        this.deltaSeconds = deltaSeconds;
         cacheControl = new CacheControl();
 
         cacheControl.setPrivate(TRUE);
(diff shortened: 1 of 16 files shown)


jcwuzoegiver:
@AndreaSuckro @mans4singh @bargenilesh @iizrailevsky @longdogz  pls review.

iizrailevsky:
@jcwuzoegiver Small import comment - otherwise looks good! 👍  :shipit:

mans4singh:
:shipit:

modules/api/src/main/java/com/intuit/wasabi/api/HttpHeader.java
```diff
@@ -59,8 +60,9 @@ public ResponseBuilder headers(int status) {
                 .cacheControl(cacheControl)
                 .header(ACCESS_CONTROL_ALLOW_ORIGIN, "*")
                 .header(ACCESS_CONTROL_ALLOW_HEADERS, "Authorization,X-Forwarded-For,Accept-Language,Content-Type")
-                .header(ACCESS_CONTROL_ALLOW_METHODS, "GET,POST,OPTIONS")
-                .header(ACCESS_CONTROL_REQUEST_METHOD, "GET,POST,OPTIONS")
+                .header(ACCESS_CONTROL_ALLOW_METHODS, "GET,POST,PUT,DELETE,OPTIONS")
+                .header(ACCESS_CONTROL_REQUEST_METHOD, "GET,POST,PUT,DELETE,OPTIONS")
```
longdogz: Is there a reason you also added PUT and DELETE here?  Wasn't mentioned as part of what this change was about?

jcwuzoegiver: @longdogz just to be consistent. because non OPTIONS (POST, GET) returns that as well.

longdogz:
OK, then looks fine to me.  :shipit:

PR #172, opened 2017-02-08
Feature/feature/assignments performance optimizations r3

bargenilesh:
- Assignments Metadata Cache - First version. As of now only used in the batch-assignment flow.
- Removed usage of user_assignment & user_assignment_by_userid tables.

@jcwuzoegiver  @mans4singh @iizrailevsky


modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
```diff
@@ -557,4 +557,36 @@ public Response getAssignmentsQueueLength() {
 
         return response;
     }
+
+
+    @POST
```
iizrailevsky: good to see both details and clear the cache API's 👍

modules/assignment/src/main/java/com/intuit/wasabi/assignment/AssignmentsModule.java
```diff
@@ -100,6 +111,42 @@ protected void configure() {
...
     }
 
+    private void bindMetadataCache(final Properties properties) {
+        Boolean metadataCacheEnabled = Boolean.parseBoolean(getProperty("metadata.cache.enabled", properties, "true"));
+        bind(Boolean.class).annotatedWith(named("AssignmentsMetadataCacheEnabled")).toInstance(metadataCacheEnabled);
+
+        if(metadataCacheEnabled) {
+            //This is a cache refresh interval, at this frequency cache will be refreshed.
+            Integer metadataCacheRefreshInterval = Integer.parseInt(getProperty("metadata.cache.refresh.interval", properties, "5"));
+            Integer metadataCacheNumberOfThreads = 1; //We want only single thread to refresh metadata cache.
```
iizrailevsky: Intentionally 1 thread for refreshing, got it. Would it be restarted if it dies for some reason?

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/AssignmentsMetadataCache.java
```diff
@@ -0,0 +1,124 @@
...
+     * @return List of experiments which are mutually exclusive to the given experiment.
+     */
+    List<Experiment.ID> getExclusionList(Experiment.ID expId);
+
+    /**
+     *
+     * @param expId
+     * @return BucketList for given experiment.
+     */
+     BucketList getBucketList(Experiment.ID expId);
```
iizrailevsky: Any reason we don't use Optional for BucketList as it's a specific class (not a List<...>)?

modules/assignment/src/main/java/com/intuit/wasabi/assignment/Assignments.java
```diff
@@ -180,6 +180,19 @@ boolean doSegmentTest(Application.Name applicationName, Experiment.Label experim
...
     /**
+     * This method is used to clear assignments metadata cache.
+     *
+     * @return True if cache is cleared successfully
+     *
+     */
+    void clearMetadataCache();
+
+    /**
+     * Clear metadata cache if it is enabled.
```
mans4singh: Should this comment be get metadata cache details and not clear metadata cache ?

modules/assignment/src/main/java/com/intuit/wasabi/assignment/AssignmentsModule.java
```diff
@@ -100,6 +111,42 @@ protected void configure() {
         LOGGER.debug("installed module: {}", AssignmentsModule.class.getSimpleName());
     }
 
+    private void bindMetadataCache(final Properties properties) {
+        Boolean metadataCacheEnabled = Boolean.parseBoolean(getProperty("metadata.cache.enabled", properties, "true"));
+        bind(Boolean.class).annotatedWith(named("AssignmentsMetadataCacheEnabled")).toInstance(metadataCacheEnabled);
+
+        if(metadataCacheEnabled) {
+            //This is a cache refresh interval, at this frequency cache will be refreshed.
+            Integer metadataCacheRefreshInterval = Integer.parseInt(getProperty("metadata.cache.refresh.interval", properties, "5"));
```
mans4singh: It might be helpful if we add units to the property name - eg:  ...interval.millis or inteval.sec.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/AssignmentsModule.java
```diff
@@ -100,6 +111,42 @@ protected void configure() {
...
+            Integer metadataCacheNumberOfThreads = 1; //We want only single thread to refresh metadata cache.
+            //This is allowed missed intervals. If cache hasn't been refreshed for more than allowed missed/buffer intervals then HealthCheck will be failed for this APP node.
+            Integer allowedStaleTime = Integer.parseInt(getProperty("metadata.cache.allowed.stale.time", properties, "360"));
```
mans4singh: Perhaps add time units to property name for clarity.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentsMetadataCacheHealthCheck.java
```diff
@@ -0,0 +1,133 @@
...
+                long maxAllowedDifferenceMS = allowedStaleTime * 60 * 1000;
+
+                //Calculate time (in milliseconds) for 2 cache refresh intervals
+                long warnDifferenceMS = 2 * metadataCacheIntervalMs;
+
+                //Calculate actual time (in milliseconds) cache has not been refreshed.
+                long diffMS = currentTimeMs - lastRefreshTimeMs;
+
+                if (diffMS > maxAllowedDifferenceMS) {
+                    //If cache has NOT been refreshed since maximum allowed stale time so fail the health check
```
iizrailevsky: Thanks for the inline comments - very helpful! 💯

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -190,7 +199,8 @@ public AssignmentsImpl(Map<String, AssignmentIngestionExecutor> executors,
         this.assignmentsRepository = assignmentsRepository;
         this.mutexRepository = mutexRepository;
         this.eventLog = eventLog;
-  
+        this.metadataCacheEnabled=metadataCacheEnabled;
```
iizrailevsky: Tidy: spaces between "=" and variables for readability :)

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -583,6 +593,128 @@ protected Assignment getAssignment(User.ID userID, Application.Name applicationN
...
+                }
+            }
+            LOGGER.debug("experimentIds for given experiment labels ({})", experimentIds);
+        } else {
+            //If allowAssignments IS NOT EMPTY means experimentBatch.labels are NOT provided.
+            //Use allowAssignments.experimentIds to populate experimentBatch.labels
+            Set<Experiment.Label> expLabels = new HashSet<>();
+            for(Experiment.ID expId:experimentIds) {
+                Experiment exp = experimentMap.get(expId);
+                if(exp!=null) {
```
iizrailevsky: Tidy: spaces between "=" :)

modules/assignment/src/test/java/com/intuit/wasabi/assignment/impl/AssignmentsMetadataCacheHealthCheckTest.java
```diff
@@ -0,0 +1,134 @@
...
+        AssignmentsMetadataCacheHealthCheck healthCheck = new AssignmentsMetadataCacheHealthCheck(metadataCache, timeService, metadataCacheInterval, allowedStaleTime, Boolean.FALSE);
+        HealthCheck.Result result = healthCheck.check();
+
+        //Verify result - It should be Healthy as cache is disabled
+        assertThat(result.isHealthy(), is(true));
+
+    }
+
+    @Test
+    public void happyPathTestCacheEnabled() {
```
iizrailevsky: Thanks for adding comprehensive tests - very helpful! 👍

modules/experiment-objects/src/main/java/com/intuit/wasabi/experimentobjects/Experiment.java
```diff
@@ -45,7 +46,7 @@
 /**
  * An experiment
  */
-public class Experiment implements Cloneable, ExperimentBase {
+public class Experiment implements Cloneable, ExperimentBase, Serializable {
```
iizrailevsky: Objects need to be Serializable to be in the EHCache, right?

iizrailevsky:
Really nice work on the caching code and unit tests @bargenilesh! A few tidy's and questions, but otherwise looks solid. 👍  :shipit:

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentMetadataCacheTimeServiceImpl.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.util.Date;
+
+/**
+ * Separate time service class so that cache functionality can be unit tested.
+ *
+ */
+public class AssignmentMetadataCacheTimeServiceImpl implements AssignmentMetadataCacheTimeService {
+    @Override
+    public Date getCurrentTime() {
```
mans4singh: Since both the impl and interface has only one method, can we place the default impl in interface ?

modules/assignment/src/main/java/com/intuit/wasabi/assignment/AssignmentsModule.java
bargenilesh: ScheduledExecutorService takes care of choosing a thread to run/execute AssignmentsMetadataCacheRefreshTask.

So, in case if current thread dies then it should create new one and use to run/execute AssignmentsMetadataCacheRefreshTask.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentsMetadataCacheImpl.java
```diff
@@ -0,0 +1,356 @@
...
+    private PagesRepository pagesRepository = null;
+    private AssignmentsMetadataCacheRefreshTask metadataCacheRefreshTask;
+    private CacheManager cacheManager;
+
+    private static final String ASSIGNMENT_METADATA_CACHE_SERVICE_NAME = "AssignmentsMetadataCache";
+
+    @Inject
+    public AssignmentsMetadataCacheImpl(@CassandraRepository ExperimentRepository experimentRepository, PrioritiesRepository prioritiesRepository,
+                                        MutexRepository mutexRepository, PagesRepository pagesRepository,
+                                        @Named("AssignmentsMetadataCacheRefreshCacheService") ScheduledExecutorService refreshCacheService,
```
mans4singh: We are using same embedded strings in at least two - perhaps can refactor and create constants.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/AssignmentsMetadataCache.java
bargenilesh: It is because of existing logic in the assignment module. I am trying to keep/not change core assignment logic but only provide what it needs from cache, in this case it needs empty BucketList object if there are no buckets.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/AssignmentsModule.java
bargenilesh: Could be, but this cache is something we would never want to be refreshed at the intervals of SECONDS or HOURS but only by MINUTES so my thoughts just keeping it at MINUTES precisions avoids any confusion or user errors.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/Assignments.java
bargenilesh: Ya thats right... Updated comment.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/AssignmentsModule.java
bargenilesh: Updated..

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentMetadataCacheTimeServiceImpl.java
bargenilesh: Prefer not to add default methods in interface as at the spirit of this feature it is intended for existing code where interface is being changed and for some reason updating all the implementations is not feasible.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentsMetadataCacheHealthCheck.java
bargenilesh: Thanks

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
bargenilesh: done

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
bargenilesh: Got it..

modules/experiment-objects/src/main/java/com/intuit/wasabi/experimentobjects/Experiment.java
bargenilesh: Yes, that's right...

modules/assignment/src/main/java/com/intuit/wasabi/assignment/AssignmentsModule.java
iizrailevsky: ack, got it

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/AssignmentsMetadataCache.java
iizrailevsky: got it

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentsMetadataCacheImpl.java
```diff
@@ -0,0 +1,356 @@
...
+
+    /**
+     *
+     * @param expId
+     * @return BucketList for given experiment.
+     */
+    @Override
+    public BucketList getBucketList(Experiment.ID expId) {
+        BucketList bucketList = null;
+        Cache cache = cacheManager.getCache(CACHE_NAME.EXPERIMENT_ID_TO_BUCKET_CACHE.toString());
```
mans4singh: General comment regarding cache values - if a value is updated, when is the value updated in the cache ?

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentsMetadataCacheImpl.java
bargenilesh: I am with you on this... But it would be odd to create constants for some of the inject names rather than for all the inject names. May be, we can work on this in separate PR to have a place to define all the Guice names/constants and use them across - that will avoid any typos and code consistency.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentsMetadataCacheImpl.java
bargenilesh: Cache values are updated at the defined frequency (default 5 minutes) by AssignmentsMetadataCacheRefreshTask

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
```diff
@@ -557,4 +557,36 @@ public Response getAssignmentsQueueLength() {
...
+    @Path("clearMetadataCache")
+    @Produces(APPLICATION_JSON)
+    @ApiOperation(value = "Clear assignments metadata cache...")
+    @Timed
+    public Response clearMetadataCache () {
```
jcwuzoegiver: @bargenilesh should only allow superadmin to execute this API.

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
```diff
@@ -557,4 +557,36 @@ public Response getAssignmentsQueueLength() {
...
+        return httpHeader.headers().entity(result).build();
+    }
+
+    /**
+     * Get the details of assignments metadata cache
+     *
+     * @return Details of assignments metadata cache - cache entities and size of each entity cache
+     */
+    @GET
+    @Path("metadataCacheDetails")
```
jcwuzoegiver: @bargenilesh should only allow superadmin to execute this API.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentMetadataCacheTimeServiceImpl.java
```diff
@@ -0,0 +1,32 @@
+/*******************************************************************************
+ * Copyright 2016 Intuit
```
jcwuzoegiver: Should be Copyright 2017

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentsMetadataCacheHealthCheck.java
```diff
@@ -0,0 +1,133 @@
+/*
+ ****************************************************************************
+ Copyright 2016 Intuit
```
jcwuzoegiver: Should be Copyright 2017

modules/assignment/src/main/java/com/intuit/wasabi/assignment/AssignmentsModule.java
jcwuzoegiver: @bargenilesh is "5" five minutes? it will be good to have a constant e.g. 5_MINUTES="5" to replace "5" to help reader to understand default is 5 minutes.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/AssignmentsMetadataCache.java
```diff
@@ -0,0 +1,124 @@
...
+     * @param pageName
+     *
+     * @return List experiments associated to the given application and page.
+     */
+     List<PageExperiment> getPageExperiments(Application.Name appName, Page.Name pageName);
+
+    /**
+     * This method is used to clear cache.
+     *
+     * @return TRUE is cache is cleared successfully else FALSE.
```
jcwuzoegiver: typo:) TRUE if cache is...

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
bargenilesh: Updated code & added respective unit test cases...

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
bargenilesh: Updated code & added respective unit test cases...

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentMetadataCacheTimeServiceImpl.java
bargenilesh: Updated code...

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentsMetadataCacheHealthCheck.java
bargenilesh: Updated code...

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/AssignmentsMetadataCache.java
bargenilesh: Corrected..

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentsMetadataCacheImpl.java
mans4singh: So if a experiment is updated, it will be refreshed for 5 minutes.  Will this cause problems if user is creating experiment with multiple buckets or is editing them ?

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentsMetadataCacheRefreshTask.java
```diff
@@ -0,0 +1,85 @@
...
+    public void run() {
+        try {
+            LOGGER.info("AssignmentsMetadataCache refresh started at = {}", timeService.getCurrentTime());
+
+            if(!refreshInProgress) {
+                //Mark that refresh has been started...
+                refreshInProgress=Boolean.TRUE;
+
+                //Refresh metadata cache
+                metadataCache.refresh();
```
jcwuzoegiver: @bargenilesh Does this method calls the underlying Casssandra repo's api to get data to refresh the cache?

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentsMetadataCacheImpl.java
```diff
@@ -0,0 +1,356 @@
...
+    @Inject
+    public AssignmentsMetadataCacheImpl(@CassandraRepository ExperimentRepository experimentRepository, PrioritiesRepository prioritiesRepository,
+                                        MutexRepository mutexRepository, PagesRepository pagesRepository,
+                                        @Named("AssignmentsMetadataCacheRefreshCacheService") ScheduledExecutorService refreshCacheService,
+                                        @Named("AssignmentsMetadataCacheRefreshInterval") Integer refreshIntervalInMinutes,
+                                        @Named("AssignmentsMetadataCacheRefreshTask") Runnable metadataCacheRefreshTask,
+                                        HealthCheckRegistry healthCheckRegistry,
+                                        @Named("AssignmentsMetadataCacheHealthCheck") HealthCheck metadataCacheHealthCheck,
+                                        CacheManager cacheManager) {
+
```
mans4singh: We can use external ehcache.xml file so that we can change the settings.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentsMetadataCacheRefreshTask.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    @Override
+    public void run() {
+        try {
+            LOGGER.info("AssignmentsMetadataCache refresh started at = {}", timeService.getCurrentTime());
+
+            if(!refreshInProgress) {
+                //Mark that refresh has been started...
+                refreshInProgress=Boolean.TRUE;
```
jcwuzoegiver: @bargenilesh is it possible the refresh will take longer than expected within the time interval? if possible, should use AtomicBoolean to make sure not more than 1 thread will come in to do the refresh.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentsMetadataCacheRefreshTask.java
```diff
@@ -0,0 +1,85 @@
...
+    private final Logger LOGGER = LoggerFactory.getLogger(AssignmentsMetadataCacheRefreshTask.class);
+
+    private AssignmentsMetadataCache metadataCache;
+    private AssignmentMetadataCacheTimeService timeService;
+
+    private Boolean refreshInProgress;
+    private Date lastRefreshTime;
+
+    @Inject
+    public AssignmentsMetadataCacheRefreshTask(AssignmentsMetadataCache metadataCache, AssignmentMetadataCacheTimeService timeService) {
```
mans4singh: We can probably use the ehcache ttl and eviction policies to remove stale values and then get fresh values on demand.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
jcwuzoegiver: q: Did you use either Eclipse/Intellij to format for all new code? It will automatically avoid this kind of comment from @iizrailevsky. Also, it will help another PR that @AndreaSuckro was working on to format all codes using either Eclipse or Intellij.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -1160,6 +1292,29 @@ public boolean doSegmentTest(Application.Name applicationName, Experiment.Label
         return queueLengthMap;
     }
 
+    /**
+     * Clear metadata cache if it is enabled.
+     */
+    @Override
+    public void clearMetadataCache() {
+        if(metadataCacheEnabled) {
```
jcwuzoegiver: @bargenilesh Does this method clear the meta cache? When there is a batch assignment call in the middle of clearing the cache, does it refresh the cache with latest meta data? Can you explain how it works?

modules/assignment/src/main/resources/assignment.properties
```diff
@@ -25,4 +25,15 @@ decision.engine.use.connection.pooling:${decision.engine.use.connection.pooling}
...
 http.proxy.port:${http.proxy.port}
-ruleCache.executor.pool.size:5
\ No newline at end of file
+ruleCache.executor.pool.size:5
+
+#Flag to enable or disable assignments metadata cache.
+metadata.cache.enabled:true
+
+#This is a cache refresh interval, at this frequency cache will be refreshed.
+metadata.cache.refresh.interval:5
```
jcwuzoegiver: @bargenilesh how many minutes for 1 interval? will be helpful to include that in the comment.

jcwuzoegiver:
@bargenilesh general comment. is it possible that client may experience different result because meta data caching is not distributed. e.g. Out of 9 app servers, server 1 may have a little stale data then server 9. Do you see any potential issues? [cc] @iizrailevsky @mans4singh

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentsMetadataCacheImpl.java
bargenilesh: This will be eventual consistent and in this case inconsistent window is up to 5 minutes.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentsMetadataCacheImpl.java
bargenilesh: Agree, added ehcache.xml file

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentsMetadataCacheRefreshTask.java
bargenilesh: Changed implementation to use AtomicBoolean...

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentsMetadataCacheRefreshTask.java
bargenilesh: That strategy will incur performance degradation on the running assignment traffic so decided to do it in background...

modules/assignment/src/main/resources/assignment.properties
bargenilesh: Added more description in comment...

bargenilesh:
This is one of the limitation (trade of with performance) of use of cache and it will be eventually consistent in up to defined interval (5 minutes).

jcwuzoegiver:
@mans4singh Looks good after you incorporated the comments. :shipit:

mans4singh:
Looks good except for the known limitations of stale cached/local data.
:shipit:

PR #175, opened 2017-02-11
Replaced string literals with constant variables: Enables compile time checking of usages

abhishekjain88:
String literals were being used for named annotations. They ran the risk of not being consistent in their usages due to typos or otherwise. To enable strict compile-time checking, it is suggested to use constants to store these literals and use the constants everywhere.


abhishekjain88:
Opened a superset at https://github.com/intuit/wasabi/pull/178. Closing this one.

PR #177, opened 2017-02-13
Feature/feature/assignments performance optimizations r3

bargenilesh:
Stability fixes after endurance test:
	-	Fix for ConcurrentModificationExcpetion while accessing BucketList
  	-	More efficient way of refreshing Assignment metadata cache to avoid frequent spike.
  	-	Added a limit of 2gb on the metadata cache
  	-	Make ehcache.xml part of conf directory	
        -	Added exception handling at AssignmentResource level to log any uncaught exception.
  	-	Some relevant assignment metadata cache logging changes.


modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentsMetadataCacheImpl.java
```diff
@@ -50,8 +52,13 @@
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
+import static com.intuit.wasabi.assignment.cache.AssignmentsMetadataCache.CACHE_NAME.APP_NAME_N_PAGE_TO_EXPERIMENTS_CACHE;
+import static com.intuit.wasabi.assignment.cache.AssignmentsMetadataCache.CACHE_NAME.APP_NAME_TO_EXPERIMENTS_CACHE;
```
mans4singh: It might be simpler to just import CACHE_NAME.*;

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentsMetadataCacheImpl.java
```diff
@@ -289,7 +259,14 @@ public BucketList getBucketList(Experiment.ID expId) {
             bucketList = (BucketList) val.getObjectValue();
         }
 
-        return isNull(bucketList)?new BucketList():bucketList;
+        return isNull(bucketList)?new BucketList():makeCopy(bucketList);
+    }
+
+    private BucketList makeCopy(BucketList bucketList) {
+        List<Bucket> iBucketList = new ArrayList<>(bucketList.getBuckets());
```
mans4singh: What's the reason for shallow copy here ?

bin/fpm.sh
```diff
@@ -93,6 +93,7 @@ for module in "$modules"; do
     modules/api/target/classes/api.properties=${home}/conf/api.properties\
     modules/api/target/generated/swagger-ui/=${home}/content/ui/dist/swagger/swaggerjson\
     modules/assignment/target/classes/assignment.properties=${home}/conf/assignment.properties\
+    modules/assignment/target/classes/ehcache.xml=${home}/conf/ehcache.xml\
```
iizrailevsky: Do we need a corresponding update on wasabi-intuit [cc: @jcwuzoegiver] to include ehcahce.xml there?

modules/assignment/src/main/resources/ehcache.xml
```diff
@@ -19,7 +19,9 @@
          xsi:noNamespaceSchemaLocation="http://ehcache.org/ehcache.xsd"
          updateCheck="true"
          monitoring="autodetect"
-         dynamicConfig="true">
+         dynamicConfig="true"
+         maxBytesLocalHeap="2G"
```
iizrailevsky: Thanks for setting a cache memory limit to ensure we don't run out of memory! 👍

iizrailevsky:
Looks good 👍  :shipit:

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentsMetadataCacheImpl.java
jcwuzoegiver: same question. Why making a copy of bucket list?

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentsMetadataCacheRefreshTask.java
```diff
@@ -49,7 +49,8 @@ public AssignmentsMetadataCacheRefreshTask(AssignmentsMetadataCache metadataCach
     @Override
     public void run() {
         try {
-            LOGGER.info("AssignmentsMetadataCache refresh started at = {}", timeService.getCurrentTime());
+            Date startTime = timeService.getCurrentTime();
+            LOGGER.debug("AssignmentsMetadataCache refresh started at = {}", startTime);
```
jcwuzoegiver: why LOGGER.debug() and then using info() later at line 65 when logging the total time it takes to refresh?

modules/assignment/src/main/resources/ehcache.xml
jcwuzoegiver: is 2G based on the type of ec2 instance we use to deploy wasabi app?

bin/fpm.sh
jcwuzoegiver: @iizrailevsky i believe so. need to do the same for wasabi-intuit's fpm.sh

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentsMetadataCacheImpl.java
jcwuzoegiver: @bargenilesh why is there ConcurrentModificationExcpetion for  BucketList? isn't the time interval 5 mins? how come multiple threads are trying to modify bucket list?

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentsMetadataCacheImpl.java
bargenilesh: Yes, you are right, it is just what IntelliJ formatter...

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentsMetadataCacheImpl.java
bargenilesh: Making a copy of BucketList to avoid ConcurrentModificationException... At some point during the assignment logic, the List<Bucket> object inside this BucketList is getting sorted, I believe that's the reason of ConcurrentModificationException. 

AssignmentImpl.selectBucket(List<Bucket> buckets) Line#1189.

Provided comments to this method...

bin/fpm.sh
bargenilesh: Already done and created a separate PR for wasabi-intuit

modules/assignment/src/main/resources/ehcache.xml
bargenilesh: 2gb is just the best guess higher limit.. We currently run app instances with 8gb of memory, out of which consistently 1.1gb is being used...

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/AssignmentsMetadataCacheRefreshTask.java
bargenilesh: In PROD, App runs at the INFO level. And I wanted to just print a single line highlighting that cache has been refreshed, so kept a single log with INFO level with the time taken to refresh.

This is based on the endurance test experiment where 3 lines were being logged at the cache refresh event, that makes it difficult to understand/track/validate behavior.

I case of any further debugging, we can explicitly enable DEBUG log for this class, which would be rare case scenario mostly in PRE-PROD or LOCAL environment.

jcwuzoegiver: ok

jcwuzoegiver:
Looks good after incorporating the feedback. :shipit:

PR #178, opened 2017-02-13
Feature/literals refactoring: Replaced string literals in named annotations with constant variables: Enables compile time checking of usages

abhishekjain88:
String literals were being used for named annotations. They ran the risk of not being consistent in their usages due to typos or otherwise. To enable strict compile-time checking, it is suggested to use constants to store these literals and use the constants everywhere.

### DIFF
--- a/modules/api/src/main/java/com/intuit/wasabi/api/ApiAnnotations.java
+++ b/modules/api/src/main/java/com/intuit/wasabi/api/ApiAnnotations.java
@@ -15,9 +15,9 @@
  *******************************************************************************/
 package com.intuit.wasabi.api;
 
-public class ApiAnnotations {
-    public static final String APPLICATION_ID = "application.id";
-    public static final String DEFAULT_TIME_ZONE = "default.time.zone";
-    public static final String DEFAULT_TIME_FORMAT = "default.time.format";
-    public static final String ACCESS_CONTROL_MAX_AGE_DELTA_SECONDS = "access.control.max.age.delta.seconds";
+public interface ApiAnnotations {
+    String APPLICATION_ID = "application.id";
+    String DEFAULT_TIME_ZONE = "default.time.zone";
+    String DEFAULT_TIME_FORMAT = "default.time.format";
+    String ACCESS_CONTROL_MAX_AGE_DELTA_SECONDS = "access.control.max.age.delta.seconds";
 }
--- a/modules/api/src/main/java/com/intuit/wasabi/api/ApiAnnotations.java
+++ b/modules/api/src/main/java/com/intuit/wasabi/api/ApiAnnotations.java
@@ -1,0 +1,23 @@
+/*******************************************************************************
+ * Copyright 2017 Intuit
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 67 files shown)


abhishekjain88:
@jcwuzoegiver @bargenilesh @iizrailevsky @mans4singh Can you please take a look at this pull request whenever you have time. It has been open for a while. Just bringing it to notice.

modules/api/src/main/java/com/intuit/wasabi/api/ApiAnnotations.java
```diff
@@ -0,0 +1,23 @@
...
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+package com.intuit.wasabi.api;
+
+public class ApiAnnotations {
```
iizrailevsky: @abhishekjain88: any reason not to use an interface instead of the class? This way you wouldn't need to explicitly put "public static final" in front of each constant as interfaces always have "public static final" fields

abhishekjain88: @iizrailevsky That actually makes a lot of sense. Let me change that and push the changes. Thanks!

abhishekjain88:
@iizrailevsky Changed all constant holder classes to interfaces.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/AssignmentsAnnotations.java
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+package com.intuit.wasabi.assignment;
+
+public class AssignmentsAnnotations {
```
iizrailevsky: Same here - possibly an interface, instead of the class?

abhishekjain88: @iizrailevsky It was already updated for all classes in my last commit after your previous comment. Can you please refresh and check.

jcwuzoegiver:
@abhishekjain88 If you are using interfaces, you may want to do "import static." e.g. import static com.intuit.wasabi.assignment.AssignmentsAnnotations.*. In this way, you just use the constant and don't have to repeat AssignmentsAnnotations. Just a suggestion. Looks good overall.

abhishekjain88:
@jcwuzoegiver Good point. Have made the changes and pushed accordingly. Can you please take a final look.

abhishekjain88:
@iizrailevsky Can you also please take a final look, so that we can merge!

jcwuzoegiver:
@abhishekjain88 looks good. :shipit:

PR #182, opened 2017-02-14
Feature/feature/assignments performance optimizations r3 - fix to disable cache functionality

bargenilesh:
- Fix to disable cache functionality

@iizrailevsky @jcwuzoegiver @mans4singh


modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/NoopAssignmentsMetadataCacheImpl.java
```diff
@@ -0,0 +1,193 @@
+/*******************************************************************************
```
jcwuzoegiver: @bargenilesh  do you have unit test for this class? just want to make sure it won't decrease code coverage.

jcwuzoegiver:
@bargenilesh looks good except 1 minor comment on unit test.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/NoopAssignmentsMetadataCacheImpl.java
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    /**
+     * @param appName
+     *
+     * @return List of experiments created in the given application.
+     */
+    @Override
+    public List<Experiment> getExperimentsByAppName(Application.Name appName) {
+        // NOOP
```
mans4singh: Can we use empty collection here rather than null ?

iizrailevsky:
:shipit:

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/NoopAssignmentsMetadataCacheImpl.java
bargenilesh: Added a new unit test

modules/assignment/src/main/java/com/intuit/wasabi/assignment/cache/impl/NoopAssignmentsMetadataCacheImpl.java
bargenilesh: This is just a NOOP implementation so returning NULL and in real implementation this method will never return NULL as core assignment logic expects NON-NULL value only...

mans4singh:
@bargenilesh - Minor comment :shipit:

PR #184, opened 2017-02-21
Feature/ehcache memory fix : Wasabi failed to start

abhishekjain88:
With the recent introduction of EHCache for caching assignment metadata, the wasabi application was failing to start on the develop branch. It was due to EHCache configured with a default "maxBytesLocalHeap" property of 2GB where as the default heap size on the JVM (if not explicitly specified) is 256MB. There is a check in the EHCache configuration code where it checks this property against the total size of the heap and if it is configured greater than the heap size, it throws the following exception:

CacheManager configuration: You've assigned more memory to the on-heap than the VM can sustain, please adjust your -Xmx setting accordingly

These changes allow configuring the application heap size and the "maxBytesLocalHeap" property at the build level so that they can be set as per requirement.

The changes have been successfully built and tested.

### DIFF
--- a/modules/assignment/src/main/resources/ehcache.xml
+++ b/modules/assignment/src/main/resources/ehcache.xml
@@ -20,7 +20,7 @@
          updateCheck="true"
          monitoring="autodetect"
          dynamicConfig="true"
-         maxBytesLocalHeap="2G"
+         maxBytesLocalHeap="${assignment.metadata.cache.maxBytesLocalHeap}"
 >
 
     <defaultCache maxEntriesLocalHeap="0" eternal="true" />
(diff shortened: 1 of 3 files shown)


abhishekjain88:
@iizrailevsky @bargenilesh Review requested.

iizrailevsky:
Looks good @abhishekjain88 - thanks for taking care of this! We might need to make corresponding changes on the wasabi-intuit branch. Please check and also get review from @jcwuzoegiver. 👍  :shipit:

bargenilesh:
@abhishekjain88 Good cache and thanks for fixing this one 👍 
Changes looks good, provided functional tests are passing :shipit: 

Also, it would be great if we make relative change (Getting memory sizes from POM) in wasabi-intuit branch.

jcwuzoegiver:
@abhishekjain88 Looks good :shipit: . As for wasabi-intuit, jvm option is already configured as follows via Maven's profile in pom.xml. 
JAVA_OPTIONS="-server -Xmx8192m\

PR #187, opened 2017-02-25
Feature/assignments performance optimizations r4

bargenilesh:
- Streamlined complete assignment module to use all the latest optimizations in single-assignment & event-recording.
- Removed traces of user_assignment & user_assignment_by_userid table.

@iizrailevsky @jcwuzoegiver @mans4singh @abhishekjain88 
@sudheerkethireddy


iizrailevsky:
👍 :shipit:

mans4singh:
Looks good :shipit:

modules/assignment-objects/src/main/java/com/intuit/wasabi/assignmentobjects/Assignment.java
```diff
@@ -245,21 +282,23 @@ public String toString() {
...
 
-        Status(boolean cacheable) {
-            this.cacheable = cacheable;
+        Status(boolean definitiveAssignment) {
+            this.definitiveAssignment = definitiveAssignment;
         }
 
-        public boolean isCacheable() {
-            return cacheable;
+        public boolean isDefinitiveAssignment() {
```
iizrailevsky: isDefinitiveAssignment() sounds a lot more appropriate than isCacheable()! 💯 :)

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
```diff
@@ -152,18 +155,18 @@ public Response getAssignment(
         Assignment assignment = getAssignment(userID, applicationName, experimentLabel, context, createAssignment,
                 ignoreSamplingPercent, null, headers);
 
-        return httpHeader.headers().entity(toMap(assignment)).build();
+        return httpHeader.headers().entity(toSingleAssignmentResponseMap(assignment)).build();
     }
 
     private Assignment getAssignment(final User.ID userID, final Application.Name applicationName,
```
abhishekjain88: Cosmetic: Suggest changing the method name to getSingleAssignment as it is just wrapping a specific assignment type.

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
```diff
@@ -152,18 +155,18 @@ public Response getAssignment(
...
                                      final Experiment.Label experimentLabel, final Context context,
                                      final boolean createAssignment, final boolean ignoreSamplingPercent,
                                      final SegmentationProfile segmentationProfile, final HttpHeaders headers) {
-        Assignment assignment = assignments.getSingleAssignment(userID, applicationName, experimentLabel, context,
-                createAssignment, ignoreSamplingPercent, segmentationProfile, headers, null);
+        Assignment assignment = assignments.doSingleAssignment(userID, applicationName, experimentLabel, context,
```
abhishekjain88: Cosmetic: Can we do away with having so many method parameters. Too many method parameter creates confusion in the code. Can we instead wrap these parameters in a POJO and pass the POJO around instead?

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
```diff
@@ -152,18 +155,18 @@ public Response getAssignment(
...
+                createAssignment, ignoreSamplingPercent, segmentationProfile, headers);
 
         // This should not happen when createAssignment == true
```
abhishekjain88: Question: Does it say that it should not happen only when createAssignment == true? Can you please explain this scenario a little bit?

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
```diff
@@ -290,10 +293,10 @@ public Response getBatchAssignments(@PathParam("applicationName")
                     headers, experimentBatch);
         }
 
-        List<Map> myAssignments = assignments.doBatchAssignments(userID, applicationName, context, createAssignment, FALSE,
-                headers, experimentBatch, null, null);
+        List<Assignment> myAssignments = assignments.doBatchAssignments(userID, applicationName, context, createAssignment, FALSE,
```
abhishekjain88: Cosmetic: Suggest changing variable name to `batchAssignments` instead of `myAssignments`.

abhishekjain88: Indentation : Please break this line.

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
```diff
@@ -290,10 +293,10 @@ public Response getBatchAssignments(@PathParam("applicationName")
...
+                headers, experimentBatch);
 
-        return httpHeader.headers().entity(ImmutableMap.<String, Object>builder().put("assignments", myAssignments).build()).build();
+        return httpHeader.headers().entity(ImmutableMap.<String, Object>builder().put("assignments", toBatchAssignmentResponseMap(myAssignments)).build()).build();
```
abhishekjain88: Indentation: Please break this line.

abhishekjain88: Cosmetic: Can we move all the string literals like `"assignments"` in this case to a string constant defined at the beginning of the class. This makes it less error prone and easy to reference and change.

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
```diff
@@ -645,11 +622,94 @@ public Response clearMetadataCache(@HeaderParam(AUTHORIZATION) @ApiParam(value =
     @Produces(APPLICATION_JSON)
     @ApiOperation(value = "Get assignments metadata cache details...")
     @Timed
-    public Response getMetadataCacheDetails(@HeaderParam(AUTHORIZATION) @ApiParam(value = EXAMPLE_AUTHORIZATION_HEADER, required = true) final String authorizationHeader) {
-        UserInfo.Username userName = authorization.getUser(authorizationHeader);
-        authorization.checkSuperAdmin(userName);
-
+    public Response getMetadataCacheDetails() {
```
abhishekjain88: Question: No authorization required for getting metadata cache details?

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
```diff
@@ -645,11 +622,94 @@ public Response clearMetadataCache(@HeaderParam(AUTHORIZATION) @ApiParam(value =
...
         return httpHeader.headers().entity(assignments.metadataCacheDetails()).build();
     }
 
+    /**
+     * Convert Assignment object collection to the List of response MAP expected by the end user.
+     *
+     * @param assignments
+     * @return
+     */
+    protected List<Map<String, Object>> toBatchAssignmentResponseMap(Collection<Assignment> assignments) {
```
abhishekjain88: Question: Why protected? Why not private?

abhishekjain88: Cosmetic: Suggest changing method name to getBatchAssignmentResponseMapsfromAssignmentCollection() to better communicate intent.

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
```diff
@@ -645,11 +622,94 @@ public Response clearMetadataCache(@HeaderParam(AUTHORIZATION) @ApiParam(value =
...
+     *  - assignment
+     *  - payload
+     *  - status
+     *
+     * @param assignment
+     *
+     * @return response map
+     *
+     */
+    protected Map<String, Object> toBatchAssignmentResponseMap(final Assignment assignment) {
```
abhishekjain88: Cosmetic: Consider renaming method name to getResponseMapfromBatchAssignmentObject to better communicate intent.

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
```diff
@@ -645,11 +622,94 @@ public Response clearMetadataCache(@HeaderParam(AUTHORIZATION) @ApiParam(value =
...
+        //Add experimentLabel for batch-assignment flow only
+        if(nonNull(assignment.getExperimentLabel())) {
+            response.put("experimentLabel", assignment.getExperimentLabel());
+        }
+
+        // Only include `assignment` property if there is a definitive assignment, either to a bucket or not
+        if (assignment.getStatus() != EXPERIMENT_EXPIRED) {
+            response.put("assignment", nonNull(assignment.getBucketLabel()) ? assignment.getBucketLabel().toString() : null);
+
+            if (nonNull(assignment.getBucketLabel())) {
```
abhishekjain88: Cosmetic: This nonNull check can be combined with the previous statement.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -198,175 +196,381 @@ public AssignmentsImpl(Map<String, AssignmentIngestionExecutor> executors,
...
+     * @param forceInExperiment
+     * @param headers           the {@link HttpHeaders} that can be used by the segmentation
+     * @param experimentBatch   the {@link ExperimentBatch} experiment batch for batch assignments
+     *                          Experiment labels are NOT PRESENT when called from AssignmentsImpl.doPageAssignments()
+     * @param pageName          the {@link com.intuit.wasabi.experimentobjects.Page.Name} the page name for the assignment
+     *                          NULL when called from AssignmentsResource.getBatchAssignments()
+     *                          PRESENT when called from AssignmentsImpl.doPageAssignments()
+     * @param allowAssignments  {@link HashMap} for each experiment whether the assignment is allowed
+     *                          NULL when called from AssignmentsResource.getBatchAssignments()
+     *                          PRESENT when called from AssignmentsImpl.doPageAssignments()
```
jcwuzoegiver: pls add comment for updateDownstreamSystems param

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -198,175 +196,381 @@ public AssignmentsImpl(Map<String, AssignmentIngestionExecutor> executors,
...
-            return nullAssignment(userID, appName, null, Assignment.Status.EXPERIMENT_NOT_FOUND);
+
+    /**
+     *
+     * Create/Retrieve assignments for a given user, application, context and
+     * given experiments (experimentBatch/allowAssignments).
+     *
+     * <p>
+     * This method is called from:
+     * 1. First directly through API call AssignmentsImpl.doBatchAssignments() => api:/v1/assignments/applications/{applicationName}/users/{userID}
```
jcwuzoegiver: @bargenilesh good comments. very helpful!

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -716,134 +789,126 @@ private void populateExperimentIdsAndExperimentBatch(Optional<Map<Experiment.ID,
...
-            throw new InvalidExperimentStateException(experiment.getID(), validStates, Experiment.State.DRAFT);
+    /**
+     * Get the experiment label based on experiment ID
+     * If metadata cache is enabled then fetch from cache else fetch from database.
+     *
+     * @param experimentID
+     * @return
+     */
+    protected Experiment.Label getExperimentLabel(Experiment.ID experimentID) {
+        if(isNull(experimentID)) return null;
```
jcwuzoegiver: did you format the code here?

jcwuzoegiver:
Looks good overall with some minor comments. :shipit:

bargenilesh:
Thanks @abhishekjain88 & @jcwuzoegiver , I have incorporated comments from you and now run functional tests and then merge to develop.

Note: Right now passing on to the comments which are on existing code, and naming, we will come back to them later, may be separate card...

abhishekjain88:
@bargenilesh Can we hold on to the merge just for today. I am still in the middle of reviewing other files.

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
bargenilesh: When createAssignment is TRUE then process will create a NEW ASSIGNMENT if it doesn't exists already.
When createAssignment is FALSE then process will NOT create a NEW ASSIGNMENT and only returns EXISTING_ASSIGNMENTs.

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
bargenilesh: Applied wasabi.xml code style formatting...

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
bargenilesh: Good idea, we may be do this in separate card..

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
bargenilesh: Yes, this is only retrieving metadata cache info and NOT changing anything so to keep API easy to use, removed authorization like queueLength etc..

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
```diff
@@ -645,11 +622,94 @@ public Response clearMetadataCache(@HeaderParam(AUTHORIZATION) @ApiParam(value =
...
+     *
+     * @return response map
+     *
+     */
+    protected Map<String, Object> toSingleAssignmentResponseMap(final Assignment assignment) {
+        Map<String, Object> response = newHashMap();
+
+        // Only include `assignment` property if there is a definitive assignment, either to a bucket or not
+        if (assignment.getStatus().isDefinitiveAssignment()) {
+            response.put("assignment", nonNull(assignment.getBucketLabel()) ? assignment.getBucketLabel().toString() : null);
```
bargenilesh: Empty string is not valid for bucket label, it throws parsing exception...

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
```diff
@@ -645,11 +622,94 @@ public Response clearMetadataCache(@HeaderParam(AUTHORIZATION) @ApiParam(value =
...
+
+            if (nonNull(assignment.getBucketLabel())) {
```
bargenilesh: it is not, because in case of definitive-assignment, even if bucket label is NULL then we still want to include 'assignment' key in the response map but not the 'payload' key...

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
```diff
@@ -645,11 +622,94 @@ public Response clearMetadataCache(@HeaderParam(AUTHORIZATION) @ApiParam(value =
...
+                response.put("payload", assignment.getPayload());
+            }
+        }
+
+        response.put("status", assignment.getStatus());
+
+        response.put("cache", assignment.getStatus().isDefinitiveAssignment());
+        if (assignment.getContext() != null) {
+            response.put("context", assignment.getContext().toString());
+        }
```
bargenilesh: We can not because then 'context' key should ONLY be part of response map if context is NOT null, with the single line it will become part of response map all the time.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
bargenilesh: Added comment for updateDownstreamSystems

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
bargenilesh: Now, applied new wasabi.xml code style..

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
bargenilesh: We can not, because in case of definitive-assignment (assignment.getStatus() != EXPERIMENT_EXPIRED), even if bucket label is NULL then we still want to include 'assignment' key in the response map but not the 'payload' key...

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
bargenilesh: Keeping them protected so that it can be override, if needed, by anyone extending wasabi-opensource to their specific wasabi-X

bargenilesh:
@abhishekjain88 Because of time sensitivity I need to merge that PR and start next steps. But Please continue your review, if there is something blocking issue found then I will create another PR.

sudheerkethireddy:
:shipit:

PR #190, opened 2017-03-02
Feature/qe automation assignments perfoptimization

sudheerkethireddy:


modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/EmptyBucketUserAssignmentForBatchAssignment.java
```diff
@@ -0,0 +1,212 @@
...
+
+            //below we are starting the experiment
+            exp.state = Constants.EXPERIMENT_STATE_RUNNING;
+            putExperiment(exp);
+
+            testExperimentsList.add(exp);
+            expList.add(exp);
+        }
+
+        // do a POST of assignments for user1234 with context set as PROD
```
bargenilesh: 'context' is deprecated (although, we haven't called it explicitly, it is) and at some point in future we will be removing it, so I would think for any new tests should be doing without 'context'

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/EmptyBucketUserAssignmentForBatchAssignment.java
```diff
@@ -0,0 +1,212 @@
...
+
+    }
+
+    /**
+     * This test case tests scenario where we do batch assignment of a user to
+     * list of experiments, delete the assigned buckets and do assignment again
+     * for the same user against the same experiments should make a new assignment
+     */
+    @Test
+    public void createBatchAssignmentAndDeleteAssignedBucketTest() {
```
bargenilesh: Should this method name be createBatchAssignmentAndEmptyAssignedBucketTest() ? different is EMPTY rather than DELETE as for bucket only following states are valid OPEN, EMPTY, CLOSED.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/EmptyBucketUserAssignmentForBatchAssignment.java
bargenilesh: On the other thought, it may be OK to keep testing context as code is changing and existing customers may be using this context and so it should work as it was working before 😄

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/assignment/SingleAssignmentContextTest.java
```diff
@@ -0,0 +1,191 @@
...
+        //create buckets for the experiment
+        List<Bucket> bucketList = BucketFactory.createBuckets(exp, 5);
+        postBuckets(bucketList);
+
+        //below we are starting the experiment
+        exp.state = Constants.EXPERIMENT_STATE_RUNNING;
+        putExperiment(exp);
+
+        //lets clear the cache and do an assignment for a user with context set to PROD
+        clearAssignmentsMetadataCache();
```
bargenilesh: No need to call clearCache() as postAssignment() internally call it..

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/EmptyBucketUserAssignmentForBatchAssignment.java
sudheerkethireddy: ok @bargenilesh

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/assignment/SingleAssignmentContextTest.java
sudheerkethireddy: removed the explicit call to clearcache and checked in the code

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/assignment/SingleAssignmentContextTest.java
```diff
@@ -0,0 +1,191 @@
...
+        Assignment assignment = postAssignment(exp, new User("user123"),"PROD");
+
+        //lets assert the status and response code
+        assertReturnCode(response, HttpStatus.SC_OK);
+        Assert.assertEquals(assignment.status,"NEW_ASSIGNMENT");
+
+        //lets repeat the assignment for same user but with different context- context set to DEV
+        clearAssignmentsMetadataCache();
```
sudheerkethireddy: removed the method call and checkedin the code

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/EmptyBucketUserAssignmentForBatchAssignment.java
sudheerkethireddy: named the functiona previously to match the sceanario but good call out renamed the method to createBatchAssignmentAndEmptyAssignedBucketTest()

bargenilesh:
Changes looks good, Provided functional tests are passing :shipit:

mans4singh:
Looks good :shipit:

iizrailevsky:
👍 :shipit:

PR #194, opened 2017-03-07
Update build.sh

esprush:
fixed UI loading on localhost:8080/dist instead of localhost:8080

### DIFF
--- a/bin/build.sh
+++ b/bin/build.sh
@@ -137,7 +137,7 @@ fi
 content=${home}/${id}/content/ui/dist
 
 mkdir -p ${content}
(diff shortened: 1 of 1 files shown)


longdogz:
@esprush , is this a problem only on certain OS's?  I've never had this as a problem when using the script on a Mac.

longdogz:
I checked Mac and our internal build environment, which is Linux.  It still worked fine on both, so I accepted this.  Thanks!

esprush:
longdogz ran into this issue in ubuntu, thanks for this great tool.

esprush:
Thanks for this great tool

Suresh P

PR #197, opened 2017-03-10
UI defaults to same host it was served on for backend API calls

longdogz:
Changed so that if the server URL that was used to load the UI is the same as the server URL of the backend, then you don’t need to change constants.json or config.js.  Instead, the UI gets the URL of the backend from the URL of the app in the browser environment.  If they need to be different, e.g., if you are running the backend in docker on localhost:8080, but you are running the UI in the foreground using “grunt serve”, then you will need to set the correct value in constants.json.

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,18 +305,6 @@ The runtime logs can be accessed executing the following command in a another sh
 % grunt build
 ```
 
-> Edit Gruntfile.js with the following change to the apiHostBaseUrlValue value, since you would be running the
-Wasabi server on localhost.
-
-```javascript
-development: {
-  constants: {
-    supportEmail: 'you@example.com',
-    apiHostBaseUrlValue: 'http://localhost:8080/api/v1'
-  }
-}
-```
-
 ```bash
 % grunt serve
 ```
--- a/modules/ui/constants.json
+++ b/modules/ui/constants.json
@@ -1,4 +1,4 @@
 {
     "supportEmail": "",
-    "apiHostBaseUrlValue": "http://localhost:8080/api/v1"
+    "apiHostBaseUrlValue": "DEFAULT"
 }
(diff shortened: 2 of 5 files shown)


longdogz:
@gittify , there needs to be a corresponding change in the README, but actually, I think it is to take out exactly the mention you were suggesting to put in.  😄 Because there is currently a setting in the Gruntfile.js such that if you do "grunt serve", it uses a backend setting of localhost:8080, but the front end is served on localhost:9000.  So it works.  If you, instead hit the UI that is being served from the docker container, e.g., localhost:8080, then that uses the DEFAULT setting.

I've updated the README.md to remove that mention, but added something about using "grunt serve:dist" in the modules/ui/README.md to explain how to make that work.  

Thanks!

modules/ui/constants.json
```diff
@@ -1,4 +1,4 @@
 {
     "supportEmail": "",
-    "apiHostBaseUrlValue": "http://localhost:8080/api/v1"
```
jcwuzoegiver: 👍  no more localhost

README.md
```diff
@@ -305,18 +305,6 @@ The runtime logs can be accessed executing the following command in a another sh
...
 ```
 
-> Edit Gruntfile.js with the following change to the apiHostBaseUrlValue value, since you would be running the
-Wasabi server on localhost.
-
-```javascript
-development: {
-  constants: {
-    supportEmail: 'you@example.com',
-    apiHostBaseUrlValue: 'http://localhost:8080/api/v1'
```
jcwuzoegiver: 👍  no more localhost

jcwuzoegiver:
Looks good. :shipit:

gittify:
:shipit:

PR #198, opened 2017-03-14
Feature/cardview resolve issues

bargenilesh:
Fix to make DBITransaction methods thread safe...

Please review: @AndreaSuckro @abhishekjain88 @jcwuzoegiver @iizrailevsky

### DIFF
--- a/modules/database/src/main/java/com/intuit/wasabi/database/impl/DBITransaction.java
+++ b/modules/database/src/main/java/com/intuit/wasabi/database/impl/DBITransaction.java
@@ -102,7 +102,7 @@ public class DBITransaction implements Transaction {
              * {@inheritDoc}
              */
             @Override
-            public Object value(Handle arg) {
+            public Object value(Handle handle) {
                 return handle.select(query, params);
             }
         });
@@ -124,6 +124,15 @@ public class DBITransaction implements Transaction {
         return insert(true, query, params);
     }
 
+    /**
+     * Make use of perform() operation to execute INSERT statement
+     *
+     * @param shouldReturnId
+     * @param query
+     * @param params
+     *
+     * @return
+     */
     private Integer insert(final boolean shouldReturnId, final String query, final Object... params) {
         try {
             return (Integer) perform(new HandleBlock() {
@@ -165,8 +174,16 @@ public class DBITransaction implements Transaction {
         });
     }
 
+    /**
+     * Execute a block (database operation) with a new handle
+     *
+     * @param block representing a database operation
+     *
+     * @return The result of a given database operation
+     *
+     */
     private Object perform(HandleBlock block) {
-        handle = getHandle();
+        Handle handle = getHandle();
         try {
             return block.value(handle);
         } catch (DBIException ex) {
@@ -178,25 +195,40 @@ public class DBITransaction implements Transaction {
 
             throw ex;
         } finally {
-            if (!inTransaction) {
-                close();
+            // Close the handle (internally this call releases the JDBC connection to the pool)
+            if (handle!=null) {
+                handle.close();
             }
         }
     }
 
+    /**
+     * Creates a new handle to perform database operation.
+     * Handle represents a connection to the database system. It is a wrapper around a JDBC Connection object.
+     *
+     * @return handle to perform database operation.
+     *
+     */
     protected Handle getHandle() {
-        if (null == handle) {
-            handle = dbi.open();
-            // Force MySQL timezone to UTC
-            handle.execute("set time_zone = \"+0:00\"");
-        }
+        //This operation reuses JDBC Connection object from the pool and then creates a new handle object.
+        Handle handle = dbi.open();
+
+        // Force MySQL timezone to UTC
+        handle.execute("set time_zone = \"+0:00\"");
 
         return handle;
     }
 
+    /**
+     * Begin a new transaction for this instance of DBITransaction.
+     *
+     * Initialize the handle of this instance of DBITransaction
+     *
+     * @throws WasabiServerException
+     */
     protected void begin() throws WasabiServerException {
         Handle h = getHandle();
-
+        handle = h;
         try {
             Connection connection = h.getConnection();
 
@@ -205,11 +237,15 @@ public class DBITransaction implements Transaction {
             throw new DatabaseException("Error obtaining database connection", e);
         }
 
-        h.begin();
+        handle.begin();
 
         inTransaction = true;
     }
 
+    /**
+     * Commits an active transaction of this instance of DBITransaction
+     *
+     */
     protected void commit() {
         if (handleIsMissing()) {
             return;
@@ -220,6 +256,10 @@ public class DBITransaction implements Transaction {
         inTransaction = false;
     }
 
+    /**
+     * Rollback an active transaction of this instance of DBITransaction
+     *
+     */
     protected void rollback() {
         if (handleIsMissing()) {
             return;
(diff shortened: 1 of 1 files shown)


jcwuzoegiver:
Looks good! :shipit:

modules/database/src/main/java/com/intuit/wasabi/database/impl/DBITransaction.java
```diff
@@ -41,9 +42,8 @@
 public class DBITransaction implements Transaction {
 
     private static final Logger LOGGER = getLogger(DBITransaction.class);
-    private Handle handle;
     private DBI dbi;
-    private boolean inTransaction = false;
+    private AtomicBoolean inTransaction = new AtomicBoolean(false);
```
abhishekjain88: What is the use of AtomicBoolean here? I didn't see it being used for any atomic operations in this class.

abhishekjain88: Also, inTransaction is being just set and the value not checked anywhere. Neither are there any getters for it so that it could be used outside the scope of this class. It makes it redundant. Can we just remove it.

modules/database/src/main/java/com/intuit/wasabi/database/impl/DBITransaction.java
```diff
@@ -178,78 +197,82 @@ private Object perform(HandleBlock block) {
 
             throw ex;
         } finally {
-            if (!inTransaction) {
-                close();
+            // Close the handle (internally this call releases the JDBC connection to the pool)
+            if (handle!=null) {
```
abhishekjain88: Let us just move this null check in the close() method and call close(handle) everywhere.

modules/database/src/main/java/com/intuit/wasabi/database/impl/DBITransaction.java
```diff
@@ -178,78 +197,82 @@ private Object perform(HandleBlock block) {
...
+        inTransaction.set(false);
     }
 
-    public void close() {
-        if (handleIsMissing()) {
-            return;
-        }
-
+    /**
+     * Close the handle (JDBC Connection) of this instance of DBITransaction and set the handle to NULL
```
abhishekjain88: We are not setting the handle to NULL anymore. Can we please remove it from the comment.

modules/database/src/main/java/com/intuit/wasabi/database/impl/DBITransaction.java
```diff
@@ -178,78 +197,82 @@ private Object perform(HandleBlock block) {
...
+     *
+     */
+    public void close(Handle handle) {
```
abhishekjain88: Possibly add a null check and an already closed check here.

abhishekjain88: Also, probably would be a good idea to wrap this close() call in a try-catch block too, logging any exceptions which occurred during close().

modules/database/src/main/java/com/intuit/wasabi/database/impl/DBITransaction.java
```diff
@@ -56,29 +56,31 @@ public DBITransaction(DBI dbi) {
...
 
             throw new DatabaseException("Unexpected exception when executing block \"" + block + "\"", e);
         } finally {
-            close();
+            close(handle);
         }
     }
 
-    private void tryRollback() {
+    private void tryRollback(Handle handle) {
```
abhishekjain88: Why do we have a separate method for doing exception handling on the rollback method. I think it is unnecessary and should be moved down to the rollback method itself.

modules/database/src/main/java/com/intuit/wasabi/database/impl/DBITransaction.java
```diff
@@ -56,29 +56,31 @@ public DBITransaction(DBI dbi) {
      */
     @Override
     public Object transaction(Block block) {
+        Handle handle = null;
         try {
-            begin();
+            handle = getHandle();
```
abhishekjain88: All the methods following this call depend on the premise that getHandle() is never going to return a null handle, otherwise, each of the following calls run the risk of throwing a NPE. Possibly a good idea to check for handle being null in getHandle() and wrapping the NPE appropriately.

modules/database/src/main/java/com/intuit/wasabi/database/impl/DBITransaction.java
```diff
@@ -178,78 +197,82 @@ private Object perform(HandleBlock block) {
...
-        Handle h = getHandle();
-
+    /**
+     * Begin a new transaction for this instance of DBITransaction.
+     *
+     * Initialize the handle of this instance of DBITransaction
+     *
+     * @throws WasabiServerException
+     */
+    protected void begin(Handle handle) throws WasabiServerException {
```
abhishekjain88: This method never throws a WasabiServerException. It is misleading. Please remove this.

modules/database/src/main/java/com/intuit/wasabi/database/impl/DBITransaction.java
bargenilesh: Removed inTransaction, now it has become absolute...

modules/database/src/main/java/com/intuit/wasabi/database/impl/DBITransaction.java
bargenilesh: As per dbi.open() handle will not be returned as NULL, it is either going to be a real value or exception...

modules/database/src/main/java/com/intuit/wasabi/database/impl/DBITransaction.java
bargenilesh: It is old code and I kept it as is.... But yes, it seems there is no need to have tryRollback()  method...

modules/database/src/main/java/com/intuit/wasabi/database/impl/DBITransaction.java
bargenilesh: Streamlined this part...

modules/database/src/main/java/com/intuit/wasabi/database/impl/DBITransaction.java
bargenilesh: DatabaseException is specific type of WasabiServerException... So it is good idea to have  WasabiServerException in throws clause.

modules/database/src/main/java/com/intuit/wasabi/database/impl/DBITransaction.java
bargenilesh: Thx, updated comment...

modules/database/src/main/java/com/intuit/wasabi/database/impl/DBITransaction.java
bargenilesh: Streamlined close() usage and modified the code as well...

iizrailevsky:
👍 :shipit:

modules/database/src/main/java/com/intuit/wasabi/database/impl/DBITransaction.java
abhishekjain88: 👍

modules/database/src/main/java/com/intuit/wasabi/database/impl/DBITransaction.java
abhishekjain88: 👍

PR #199, opened 2017-03-15
Feature/code coverage

iizrailevsky:
Standardized jacoco.exec and jacoco.xml files to be writtern to [WS]/modules/\<moduleName]>/target/jacoco.\*, instead of [WS]/modules/\<moduleName\>/modules/target/jacoco.\*

This is to make sure both Travis CI/Coveralls and Jenkins/Sonar find jacoco files successfully


abhishekjain88:
👍 :shipit:

PR #200, opened 2017-03-15
Feature/ui make zip from feature

longdogz:


abhishekjain88:
@longdogz Looks like it has conflicting changes to the jenkins.sh file, with the changes from the superadmin branch. Can you please take care of that and repost for review.

bin/wasabi.sh
```diff
@@ -42,6 +43,7 @@ options:
   -p | --profile [ profile ]             : profile; default ${profile_default}
   -e | --endpoint [ host:port ]          : api endpoint; default: ${endpoint_default}
   -v | --verify [ true | false ]         : verify installation configuration; default: ${verify_default}
+  -v | --buildtests [ true | false ]     : perform tests after build; default: ${buildtests_default}
```
abhishekjain88: @longdogz you should have a different shorthand from "-v" for this option otherwise it conflicts with the shorthand for --verify. Possibly "-t" ?

longdogz:
Nice catch, @abhishekjain88 , although it was only that in the usage message.  Turns out it was actually -t in the implementation (at the bottom of the file).  I've fixed the message though.

PR #201, opened 2017-03-16
Feature/ui change terminate icon

longdogz:


modules/ui/app/scripts/services/UtilitiesFactory.js
```diff
@@ -1002,16 +1002,22 @@ angular.module('wasabi.services').factory('UtilitiesFactory', ['Session', '$stat
...
                         break;
                     case 'terminated':
                         stateChange = 'terminate';
+                        title = 'Permanently Terminate Experiment';
                         break;
                 }
-                DialogsFactory.confirmDialog('Are you sure you want to ' + stateChange + ' the experiment ' + experiment.label + '?', 'Confirm State Change',
+                var msg = 'Are you sure you want to ' + stateChange + ' the experiment ' + experiment.label + '?';
+                if (state.toLowerCase() === 'terminated') {
+                    msg = 'Are you sure you want to <span style="font-weight: bold;">PERMANENTLY TERMINATE</span> the experiment ' + experiment.label + '?';
```
gittify: this double confirmation is a good idea 👍 
Will there be a help icon explaining a bit more about permanently terminate state...?

longdogz: There is currently a description of the terminate state on help accessible from the experiments list (icon next to State in the headers of the list).  Do you think that is sufficient?  (cc @calvinwma )?  Or do you think we should also put a help icon near the control on the Details dialog?

gittify: Just thought its easier on the dialog box but its optional....we can always add it later too

modules/ui/app/scripts/controllers/DialogModalCtrl.js
```diff
@@ -2,12 +2,12 @@
 
 angular.module('wasabi.controllers')
     .controller('DialogModalCtrl',
-        ['$scope', '$modalInstance', 'options', '$timeout',
-            function ($scope, $modalInstance, options, $timeout) {
+        ['$scope', '$modalInstance', 'options', '$timeout', '$sce',
```
abhishekjain88: Suggest using the expanded version for the variable name "sce" just to make it more explanatory while reading code.

sudheerkethireddy:
looks good :shipit:

modules/ui/app/scripts/controllers/DialogModalCtrl.js
longdogz: That's a built-in variable from Angular JS (1.x).  I had to look up what it stands for.  😄 I don't believe they have implemented it as a longer named variable, but I'll add a comment, "Strict Contextual Escaping".

PR #203, opened 2017-03-18
Cassandra docker image refresh with latest migration scripts

abhishekjain88:
Currently, the migration scripts present in the wasabi_keyspace and wasabi_migration docker image are from an old version and uploaded into the felixgao docker images. They are reused over and over again despite any changes or new additions in the migration scripts.

This change removes the dependency on the felixgao version for docker images. This allows the user of this repository to always get the latest migration scripts in the cassandra dockers the first time it clones the repository and tries to start wasabi.

Also it introduces a new option "-m" in the wasabi.sh and container.sh scripts to force refresh of the cassandra docker images in case changes to migration scripts or introduction of new migration scripts needs to be reflected.

Also, changed wasabi docker image to be built each time wasabi is started in order to get the latest build.

### DIFF
--- a/bin/container.sh
+++ b/bin/container.sh
@@ -35,7 +35,7 @@ usage: `basename ${0}` [options] [commands]
 
 options:
   -v | --verify [ true | false ]         : verify installation configuration; default: ${verify_default}
-  -m | --migration [ true | false ]      : refresh migration scripts; default: ${migration_default}
+  -m | --migration [ true | false ]      : refresh cassandra migration scripts; default: ${migration_default}
   -s | --sleep [ sleep-time ]            : sleep/wait time in seconds; default: ${sleep_default}
   -h | --help                            : help message
 
--- a/bin/container.sh
+++ b/bin/container.sh
@@ -271,7 +271,7 @@ exec_commands_project() {
 }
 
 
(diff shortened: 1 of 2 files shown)


iizrailevsky:
Added a couple of questions/comments. Looks good overall! If tested properly, let's :shipit:

bin/container.sh
```diff
@@ -256,20 +271,23 @@ exec_commands_project() {
 }
 
 
-optspec=":f:p:v:s:h-:"
+optspec=":f:p:v:m::s:h-:"
```
abhishekjain88: Nice catch. 👍  No specific reason. It is a typo but seemed to work fine functionally despite that. Removed now.

bin/container.sh
```diff
@@ -34,6 +35,7 @@ usage: `basename ${0}` [options] [commands]
 
 options:
   -v | --verify [ true | false ]         : verify installation configuration; default: ${verify_default}
+  -m | --migration [ true | false ]      : refresh migration scripts; default: ${migration_default}
```
abhishekjain88: Makes sense @iizrailevsky . Changed it to refresh cassandra migration.

jcwuzoegiver:
Looks good! :shipit: provided it works.

PR #207, opened 2017-03-22
user id varchar 200 and api returns hypothesis and results

gittify:


longdogz:
I'm not sure what "user id archer 200" means, but maybe it's a typo about increasing the allowed length of user IDs to 200 chars?  😄  I see that in the changes, and they look OK (as far as I can understand).  Also, I see the change to add the missing get of the hypothesis and results.  Looks good to me!  :shipit:

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/impl/ExperimentHelper.java
```diff
@@ -81,6 +81,8 @@ public static Experiment makeExperiment(
         experimentObject.setUserCap(experimentPojo.getUserCap());
         experimentObject.setCreatorID(experimentPojo.getCreatorId());
         experimentObject.setRuleJson(convertRuleToJson(experimentPojo.getRule()));
+        experimentObject.setHypothesisIsCorrect(experimentPojo.getHypothesisIsCorrect());
+        experimentObject.setResults(experimentPojo.getResults());
```
longdogz: I guess the one question I would have is just to make sure that whatever this technology you guys use to put things in an object and then eventually return JSON handles things like quotes and curly braces, basically the things that are special to JSON?  I assume you must, but if that is a special step you need to take, you'd want to do it here.  Because what if my results consist of:  Our so-called "experiment", which tests curly braces ({ and }) isn't that conclusive.

😄

iizrailevsky:
Looks good! 👍 :shipit:

PR #208, opened 2017-03-22
Fixing a formatting problem from the fix to avoid accidental terminat…

longdogz:
…ions.


longdogz:
I'm going to go ahead and merge this pull request with only one :shipit: as it is such a simple change and only in CSS.

PR #210, opened 2017-03-24
Commented out deleteLastSuperAdmin test with a TODO to make integration tests independent

abhishekjain88:
The deleteLastSuperAdmin test is failing because it assumes that admin is the only superadmin remaining in the table which is a bad assumption, as that might not be the case always.

We might need to do a cleansing activity on the integration tests, so that they can be made as independent as possible - possibly by using dedicated keyspaces but I guess that will be a longer term effort.

Simulating this kind of a situation where "admin" is the last superadmin present requires deleting all other superadmins which might not be desirable by other integration tests which depend on the same table/keyspace.

So, I suggest the best possible thing to do right now is to just deactivate that test for now with a TODO  - I had done the same in wasabi-intuit before merging the changes as a much greater refactoring might be needed on the integration tests.

### DIFF
--- a/modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/IntegrationAuthorization.java
+++ b/modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/IntegrationAuthorization.java
@@ -316,10 +316,15 @@ public class IntegrationAuthorization extends TestBase {
         Assert.assertEquals(isWasabiSuperadmin, false);
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


abhishekjain88:
@iizrailevsky @sudheerkethireddy @gittify Requesting review.

iizrailevsky:
Makes sense! 👍 :shipit:

bargenilesh:
Looks good, :shipit:

PR #213, opened 2017-03-28
Batch insert into STAGING table...

bargenilesh:
Batch insert into STAGING table to facilitate downstream-system queue draining on the queue FULL event...

@iizrailevsky @ajain11 @ndoshi @jcwuzoegiver

### DIFF
--- a/modules/assignment/src/main/resources/assignment.properties
+++ b/modules/assignment/src/main/resources/assignment.properties
@@ -27,6 +27,14 @@ http.proxy.host:${http.proxy.host}
 http.proxy.port:${http.proxy.port}
 ruleCache.executor.pool.size:5
 
+#Specify the number of messages to be drained on the event of queue full (executor.max.queue.size).
+#Default value is 10_000
+drain.queue.by=10000
+
+#Specify the batch size to be used to drain queue
+#Default value is 1_000
+drain.queue.batch.size=1000
+
 #Flag to enable or disable assignments metadata cache.
 metadata.cache.enabled:true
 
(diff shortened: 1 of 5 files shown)


modules/assignment/src/main/resources/assignment.properties
```diff
@@ -27,6 +27,14 @@ http.proxy.host:${http.proxy.host}
 http.proxy.port:${http.proxy.port}
 ruleCache.executor.pool.size:5
 
+#Specify the number of messages to be drained on the event of queue full (executor.max.queue.size).
+#Default value is 10_000
+drain.queue.by=10000
```
iizrailevsky: These must be the default values just in case we need to implement queue draining functionality in wasabi open source.

iizrailevsky:
Looks good! Thanks for adding the unit tests 👍  :shipit:

modules/assignment/src/main/resources/assignment.properties
bargenilesh: @abhishekjain88 @iizrailevsky Yes, for consistency, I have added these placeholder configs in wasabi-open source which could be used for queue draining functionality in future...

abhishekjain88: Agreed @bargenilesh @iizrailevsky I figured that and that is the reason why I said that you should create appropriate bindings in the code with named annotations so that when it needs to be extended, the property values are already available to be used with named annotations.

gittify:
:shipit:

abhishekjain88:
:shipit:

PR #215, opened 2017-04-04
Fix bug that superadmin menu choice is shown to admins

longdogz:
Fixing bug that if you are not a superadmin, but you are an admin, you see the Superadmins menu choice.  It doesn’t work, because the API is still access controlled.


abhishekjain88:
:shipit:

PR #217, opened 2017-04-04
added util method 

sudheerkethireddy:
This PR is to handle the offset that we send to cardView API

### DIFF
--- a/modules/functional-test/src/main/java/com/intuit/wasabi/tests/library/util/Constants.java
+++ b/modules/functional-test/src/main/java/com/intuit/wasabi/tests/library/util/Constants.java
@@ -106,4 +106,26 @@ public class Constants {
 
     public static final String ROLE_LIST = "roleList";
 
+    /**
+     * This enum class holds the timeZones of various regions..added a few timezones
+     * feel free to add your timeZone if it is not there...
+     */
+    public static  enum TimeZoneEnum {
+        USA_PACIFIC("America/Los_Angeles"),
+        USA_MOUNTAIN("America/Denver"),
+        USA_CENTRAL("America/Chicago"),
+        USA_EAST("America/Detroit"),
+        USA_ALASKA("America/Anchorage") ;
+
+        private String timeZone;
+
+        TimeZoneEnum(String timeZone) {
+            this.timeZone = timeZone;
+        }
+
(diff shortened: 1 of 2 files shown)


modules/functional-test/src/main/java/com/intuit/wasabi/tests/library/util/Constants.java
```diff
@@ -106,4 +106,26 @@
 
     public static final String ROLE_LIST = "roleList";
 
+    /**
+     * This enum class holds the timeZones of various regions..added a few timezones
+     * feel free to add your timeZone if it is not there...
+     */
+    public static  enum TimeZoneEnum {
```
bargenilesh: @sudheerkethireddy How & where this TimeZoneEnum class is used?

sudheerkethireddy: this one will be used for the card view API @bargenilesh where we send the offset as part of the request

modules/functional-test/src/main/java/com/intuit/wasabi/tests/library/util/Constants.java
```diff
@@ -106,4 +106,26 @@
...
+        USA_PACIFIC("America/Los_Angeles"),
+        USA_MOUNTAIN("America/Denver"),
+        USA_CENTRAL("America/Chicago"),
+        USA_EAST("America/Detroit"),
+        USA_ALASKA("America/Anchorage") ;
+
+        private String timeZone;
```
bargenilesh: @sudheerkethireddy Instead of string timezone USA_ALASKA("America/Anchorage") , can you store TimeZone object itself; this will save creation of multiple TimeZone objects... e.g.g USA_ALASKA(new TimeZone("America/Anchorage"));

iizrailevsky: Agreed - we should lookup actual ZoneId objects: https://docs.oracle.com/javase/tutorial/datetime/iso/timezones.html

sudheerkethireddy: ack: @iizrailevsky  @bargenilesh

PR #222, opened 2017-04-06
Null dereference check on EventEnvelopePayload toJson utility

abhishekjain88:
We were seeing null pointer exceptions on the export DB internally on Wasabi's intuit extension which were caused by unsafe dereferencing of null objects while using the toJson utility on the event envelope. This change addresses the fix.

There is a parallel PR on wasabi intuit as well.

### DIFF
--- a/modules/event-objects/src/main/java/com/intuit/wasabi/eventobjects/EventEnvelopePayload.java
+++ b/modules/event-objects/src/main/java/com/intuit/wasabi/eventobjects/EventEnvelopePayload.java
@@ -88,18 +88,27 @@ public class EventEnvelopePayload implements EnvelopePayload {
         JSONObject eventJson = new JSONObject();
 
         eventJson.put("messageType", MessageType.EVENT.toString());
-        eventJson.put("applicationName", applicationName.toString());
-        eventJson.put("experimentLabel", experimentLabel.toString());
-        eventJson.put("userID", assignment.getUserID().toString());
-        eventJson.put("bucketLabel", assignment.getBucketLabel().toString());
+        eventJson.put("applicationName", applicationName != null ? applicationName.toString(): "");
+        eventJson.put("experimentLabel", experimentLabel != null ? experimentLabel.toString(): "");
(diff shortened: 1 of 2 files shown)


iizrailevsky:
Just few tidy comments. Once resolved and integration tests pass -> :shipit:

modules/event-objects/src/main/java/com/intuit/wasabi/eventobjects/EventEnvelopePayload.java
```diff
@@ -88,18 +88,25 @@ public String toJson() {
...
-        eventJson.put("eventPayload", event.getPayload());
-        eventJson.put("value", event.getValue());
+        eventJson.put("experimentID",
+                assignment != null && assignment.getExperimentID() != null ?
+                        assignment.getExperimentID().toString(): null);
+        eventJson.put("context",
+                assignment != null && assignment.getContext() != null ? assignment.getContext().toString(): null);
+        eventJson.put("epochTimestamp",
+                event != null && event.getTimestamp() != null ? event.getTimestamp().getTime(): null);
+        eventJson.put("eventType", event != null && event.getType() != null ? event.getType() + "": "null");
```
bargenilesh: Minor: You probably don't need "&& event.getType() != null" & "&& event.getName() != null"

bargenilesh:
Except a very minor comment, looks good... Provided integration tests pass :shipit:

modules/event-objects/src/main/java/com/intuit/wasabi/eventobjects/EventEnvelopePayload.java
abhishekjain88: Thanks @bargenilesh for pointing that out. I didn't realize that you could concatenate with a "null" string.

abhishekjain88: Taken care in latest commit!

PR #223, opened 2017-04-09
Update README.md

iizrailevsky:
Removed the Mac OS X only note from the README as we now support other platforms, such as Linux and Windows.

Also added ":" after 'A/B Testing Overview' for consistency at the top


longdogz:
:shipit:

PR #224, opened 2017-04-10
Move Cassandra cql scripts to resources

jcwuzoegiver:
Move cql scripts under resources so it will be included in the jar so all release jars will include the cql scripts.


jcwuzoegiver:
@bargenilesh @iizrailevsky pls review

abhishekjain88:
:shipit:

abhishekjain88:
@jcwuzoegiver [cc @bargenilesh @iizrailevsky ] - We missed out on updating the cqlsh docker script with the new location of the mutation scripts. We can either reopen the same PR or fix it as a separate PR. I can do the change if you want.

abhishekjain88:
Added PR here: https://github.com/intuit/wasabi/pull/226

gittify:
as far as it is tested and works :shipit:

PR #233, opened 2017-04-28
Feature/experiment tags

longdogz:


modules/repository-datastax/db/mutation/V035__Create_ExperimentTag_table.cql
```diff
@@ -0,0 +1,5 @@
+create table experiment_tag (
```
bargenilesh: New location for cassandra delta is "modules/repository-datastax/src/main/resources/com/intuit/wasabi/repository/impl/cassandra/migration"

modules/repository-datastax/db/mutation/V036__Alter_Experiment_Add_Tags.cql
```diff
@@ -0,0 +1 @@
+alter table experiment ADD tags set<text>;
```
bargenilesh: New location for cassandra deltas: modules/repository-datastax/src/main/resources/com/intuit/wasabi/repository/impl/cassandra/migration

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/accessor/ExperimentTagAccessor.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Accessor for the tags associated with Experiments.
+ */
+@Accessor
+public interface ExperimentTagAccessor {
+
+    @Query("select * from experiment_tag where app_name IN ?")
```
bargenilesh: IN query is not efficient in the Cassandra... Let's use ASYNC implementation to retrieve tags for multiple applications. Please refer to the ExperimentAccessor.asyncGetExperimentById() and it's usages.

AndreaSuckro: @bargenilesh is this also true for such a small table? It only contains one row per application (we currently have 278 and we only store those which have tags associated with them) and usually users will only have access to a handful of those. It feels like having now one call for each application could slow things down, but maybe I am getting something wrong. I am also not sure about the partitioning in general: if I want the complete table (since it is so small) to be replicated across the nodes, how would I do that? Or does that make no sense? Thanks!

modules/repository-datastax/db/mutation/V036__Alter_Experiment_Add_Tags.cql
AndreaSuckro: Thanks - fixed it :)

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/accessor/ExperimentTagAccessor.java
AndreaSuckro: I replaced it now and would appreciate your feedback :)

modules/repository-datastax/src/main/resources/com/intuit/wasabi/repository/impl/cassandra/migration/V035__Create_ExperimentTag_table.cql
```diff
@@ -0,0 +1,5 @@
+create table experiment_tag (
+  app_name text,
+  tags set<text>,
```
gittify: does this table not need an experiment name since experiment_tag ?

AndreaSuckro: This table stores the experiment tags per application, that's why I called it experiment_tag :)

bargenilesh:
@AndreaSuckro For backend changes, good job 👍 
 :shipit: provided unit & integration tests pass...

modules/api/src/main/java/com/intuit/wasabi/api/ApplicationsResource.java
```diff
@@ -373,4 +376,48 @@ public Response getPagesAndAssociatedExperimentsForApplication(
...
+            @ApiParam(value = EXAMPLE_AUTHORIZATION_HEADER, required = true)
+            final String authorizationHeader) {
+        try {
+
+            // get the for the user visible Applications
+            List<Application.Name> allApplications = experiments.getApplications();
+            Set<Application.Name> allowed = new HashSet<>();
+            UserInfo.Username userName = authorization.getUser(authorizationHeader);
+
+            for (Application.Name applicationName : allApplications) {
```
abhishekjain88: @AndreaSuckro This loop will end up making a cassandra query for each application (for all the applications in the system irrespective of what the user is authorized to) which in my opinion is highly inefficient. Would suggest using the `getUserPermissionsList()` method instead of `checkUserPermissions()` which would fetch all the user permissions for all applications that it is authorized to view with the application names. You can probably then also remove the `getApplications()` call which will be redundant in that case.

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/impl/CassandraExperimentRepository.java
```diff
@@ -1194,4 +1234,35 @@ public void createApplication(Application.Name applicationName) {
...
+     */
+    @Override
+    public Map<Application.Name, Set<String>> getTagListForApplications(Collection<Application.Name> applicationNames) {
+
+        LOGGER.debug("Retrieving Experiment Tags for applications {}", applicationNames);
+
+        try {
+            List<ListenableFuture<Result<ExperimentTagsByApplication>>> futures = new ArrayList<>();
+
+            for (Application.Name appName : applicationNames) {
```
abhishekjain88: @AndreaSuckro I understand @bargenilesh suggested that the IN query in Cassandra is not efficient but probably @bargenilesh can confirm whether it is true for small number of rows as well (as you had rightly asked earlier). If it is inefficient for such small rows as well, then probably instead of hitting cassandra with a query for each application, it would be better to just fetch all the rows (which is expected to be in the order of 100s or worst case 1000s which should still be ok) and do an in-memory filtering (within our code) for applications we need. That would be much faster and cleaner.

abhishekjain88: We could even maintain a cache (reusing the experiment metadata cache) for all applications with tags.

modules/api/src/main/java/com/intuit/wasabi/api/ApplicationsResource.java
bargenilesh: @abhishekjain88 Agree, it would be smart to use getUserPermissionsList() first to get the list of user permissions (Applications & roles in those applications) and then get the tags for only applicable applications...

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/impl/CassandraExperimentRepository.java
bargenilesh: @abhishekjain88 Basically, in Cassandra, using IN query is ANTI design pattern and no matter number of rows it will be comparably better to use parallel point queries instead of single IN query... Plus, having IN query call available in Repository is risky if in future someone use that in the transaction-API calls (e.g. assignments, event-recording etc). 

About other suggestions, they may be better but would not be worth doing it... Considering that #1 this is admin-API call - low volume. #2 As long as it is a point query Cassandra should able to handle it in efficient way.

Regarding cache use, it is NOT good idea to use any kind of APP level caching for admin-API call..

abhishekjain88: @bargenilesh [cc @AndreaSuckro ] I agree with what you said regarding the IN queries. Also, I agree on the second part too where it might be over-engineering this thing when the API expects low volume and the queries to cassandra are made in parallel using async mode so that network I/O does not bottleneck the overall responsiveness of the API even in case of reasonable number of experiments.

Also, putting this into the cache might make the admin API a little more responsive but I guess we might be able to live with the current responsiveness for now.

@bargenilesh Appreciate your response. Thanks!

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/impl/CassandraExperimentRepository.java
```diff
@@ -402,6 +417,28 @@ protected Experiment internalGetExperiment(Application.Name appName,
...
+     * @param applicationName the Application for which the tags should be added
+     * @param tags            list of tags for the experiment
+     */
+    protected void updateExperimentTags(Application.Name applicationName, Set<String> tags) {
+
+        if (tags == null)
+            tags = Collections.EMPTY_SET;
+
+        // update the tag table, just rewrite the complete entry
+        List<ExperimentTagsByApplication> listAllTags = experimentAccessor.getAllTags(applicationName.toString()).all();
```
abhishekjain88: @AndreaSuckro Why are we retrieving the tags from the `experiment` table? Could we not just call `update()` on  `ExperimentTagAccessor` instead with the newly passed `tags` set in the method parameter? It will also avoid the race condition of tags getting overwritten by 2 different calls at the same time due to the rewrite of the complete entry. Let me know what you think on this.

abhishekjain88: @AndreaSuckro I can take a guess that you might have done this to update the `experiment_tag` table if it is not consistent with the tags in the `experiment` table but it should ideally not happen anytime as tags get updated in both tables when new experiment is created or an existing experiment is modified. Is there any other flow that might update tags?

Also, in what scenario would you want to remove tags from an application (from a product requirement) because rewriting makes sense in that case too? If you want to handle deletes of tags on application, then this code in its current state does not handle that too in my opinion.

Pardon my guesses! :) I was just trying to put all information upfront because of our timezone difference. :)

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/impl/CassandraExperimentRepository.java
```diff
@@ -402,6 +417,28 @@ protected Experiment internalGetExperiment(Application.Name appName,
...
 
     /**
+     * Updates the tags for a given Application.
+     *
+     * @param applicationName the Application for which the tags should be added
+     * @param tags            list of tags for the experiment
+     */
+    protected void updateExperimentTags(Application.Name applicationName, Set<String> tags) {
+
+        if (tags == null)
```
abhishekjain88: Can we not cut the loop here itself and not proceed if `tags` is null or empty? Am I missing something @AndreaSuckro ?

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/impl/CassandraExperimentRepository.java
AndreaSuckro: @abhishekjain88 I still feel like replicating the whole table to all nodes and then doing an IN query is the best solution, but I agree that it is not really worth spending more time if this is a solution everyone can live with.

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/impl/CassandraExperimentRepository.java
AndreaSuckro: @abhishekjain88 : no worries, thanks for your input! I am not sure if I understood your concern though. Let me explain my reasoning. The reason for this code was my following thought (correct me if I am wrong): imagine you have several experiments for one application. Two of them have the following tags exp1 - [a,b], exp2 - [a,b,c]. The index table that holds the mapping between applications and their tags is now holding the tags that are present for that application app1 : [a,b,c,d,f,....]. What happens in the case that exp2 has a label removed e.g. b? After the experiment was updated, the index table has to be updated too. Now it would be not too smart to just remove the tag 'b' from the index table, since exp1 is still holding that tag. Instead of having now a code that checks for all combinations of possible updates and strange edge-cases, I decided to let the experiment table be the ground truth for the values of the index. You wrote that this should ideally not happen, but this method is exactly there for filling the `experiment_tag` table. Does that make sense?

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/impl/CassandraExperimentRepository.java
AndreaSuckro: This is necessary in the case when tags should be deleted from the application, but maybe have a look at my other comment to this method for clarification.

modules/api/src/main/java/com/intuit/wasabi/api/ApplicationsResource.java
AndreaSuckro: @abhishekjain88 Cool - I didn't know about this function, I bet there are lot of places in the code that could also benefit from it. Please confirm that I used it correctly :)

modules/ui/app/scripts/controllers/ApplicationsCtrl.js
```diff
@@ -60,7 +60,7 @@ angular.module('wasabi.controllers').
             $scope.loadApplications();
 
             $scope.openApplicationModal = function (application) {
-                var modalInstance = $modal.open({
+                var modalInstance = $uibModal.open({
```
longdogz: I guess I would say no.  The class name and the associated "$ var" was changed to make the global name space a little less likely to have conflicts (I think) and to reflect that these things are related to the Bootstrap UI (hence "uib").  But the local var is still a reference to the "modal dialog instance".

modules/ui/app/scripts/controllers/BucketAssignmentModalCtrl.js
```diff
@@ -68,7 +68,7 @@ angular.module('wasabi.controllers')
                         // Editing the bucket, don't want to be able to edit the allocation here.
                         isEditFlag = false;
                     }
-                    var modalInstance = $modal.open({
+                    var modalInstance = $uibModal.open({
```
longdogz: Same answer.

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/impl/CassandraExperimentRepository.java
abhishekjain88: That makes a lot of sense @AndreaSuckro . In fact, it was one of my guesses (if you refer to my comment above ;) ) on the reasoning behind doing this and that is why I asked that whether we really want to delete the tag from the application tag index table as I presume that the index table is for the autocomplete feature in the UI (from a product perspective). Autocomplete in my opinion should show history of tags created, irrespective of the fact that some experiments might have been untagged from a particular tag. This is something that we might want to check with @longdogz or @calvinwma. If it is not required to remove tags, then we can get rid of the rewriting altogether.

The reason I am against the rewriting logic is that it is not an atomic operation and might result in race conditions. For example, 2 users are logged in into the admin interface. One of them is creating, deleting, updating a tag for an experiment and the other one is also either creating, deleting, updating a tag for another experiment. The first user reads state A of tags in the experiment table and the second user also reads state A of tags. Now, the first user updates the tag state A -> B where as the second user updates the tag state A -> C. One of the state changes overwrites the other and the latter gets lost permanently until there is a new update on tags called when your method updates the indices again. Does that make sense or did I confuse you? :)

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/impl/CassandraExperimentRepository.java
abhishekjain88: @AndreaSuckro It is a little non-intuitive to figure the deletes case from the code here. Would suggest either adding a comment explicitly saying that it handles the delete tag case.

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/impl/CassandraExperimentRepository.java
abhishekjain88: Also @AndreaSuckro , if we really want to support the `delete tag` use case on the application-tag index table, I think a safer and more optimal way of doing that would be to not use a set of tags in one single row but instead have another column for `experiment id`. So, your experiment_tag table would have the `app_name` as the partition key and 2 other columns which are `experiment id` and `tags`. `experiment_id` would be your clustering key. When you update a tag to the experiment, you update the `experiment` table as you are doing right now and you update the `experiment_tag` table in the corresponding row of that `experiment_id`. When you want to retrieve the list of tags for an application, you just query the `experiment_tag` table using the `app_name` as the key and it will return you a result set of all rows containing different experiments and tags corresponding to that `app_name`. You can then merge the list of tags in the code and return. 

That way, you will only be affecting the row for a particular experiment at a time, using atomic operations on the set of tags (either add or delete) and you avoid querying the `experiment` table for updating the `experiment_tag` table.

Does that sound like a reasonable approach?

Apologies for suggesting it too late in the review cycle but I guess it is important to do it the right way. If you do not understand anything, feel free to schedule a BlueJeans session with me if you want.

modules/api/src/main/java/com/intuit/wasabi/api/ApplicationsResource.java
```diff
@@ -395,19 +396,13 @@ public Response getAllExperimentTags(
             @ApiParam(value = EXAMPLE_AUTHORIZATION_HEADER, required = true)
             final String authorizationHeader) {
         try {
-
-            // get the for the user visible Applications
-            List<Application.Name> allApplications = experiments.getApplications();
-            Set<Application.Name> allowed = new HashSet<>();
             UserInfo.Username userName = authorization.getUser(authorizationHeader);
+            Set<Application.Name> allowed = new HashSet<>();
```
abhishekjain88: @AndreaSuckro Thanks for making these changes quickly! You could possibly rename this variable to `authorizedApplications` for better readability.

modules/api/src/main/java/com/intuit/wasabi/api/ApplicationsResource.java
```diff
@@ -395,19 +396,13 @@ public Response getAllExperimentTags(
...
-                try {
-                    authorization.checkUserPermissions(userName, applicationName, READ);
-                    allowed.add(applicationName);
-                } catch (AuthenticationException ignored) {
-                    LOGGER.trace("ignoring authentication exception", ignored);
-                }
+            // get the for the user visible Applications
+            List<UserPermissions> authorized = authorization.getUserPermissionsList(userName).getPermissionsList();
+            for (UserPermissions perm : authorized) {
+                allowed.add(perm.getApplicationName());
```
abhishekjain88: @AndreaSuckro Just a question here! I saw that we were explicitly checking for `READ` permission in your old code but we do not check for the permission type now. It makes sense though to not check the permission type as I presume `READ` would be the lowest permission type and if there is a record for the user in the permissions table with the given app, we can safely consider the user authorized for atleast `READ` permissions. Correct?

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/impl/CassandraExperimentRepository.java
AndreaSuckro: Hm so first I think the delete makes sense, because you use it in the filtering completion and that only makes sense for existing tags. It also allows Users to make mistakes and not clutter their applications history forever.

And I understand your concerns concerning the dirty read/write and I think your suggestion is good. I just don't feel it to be so crucial for this feature. In the unlikely event of a simultaneous change to the tags of different experiments, the worst thing to happen is a wrong suggestion in the auto-complete on the search. But it can be changed of course - it will just take a bit of time :D

modules/api/src/main/java/com/intuit/wasabi/api/ApplicationsResource.java
AndreaSuckro: That's how I understood it at least :)

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/impl/CassandraExperimentRepository.java
abhishekjain88: @AndreaSuckro I do understand that it might happen in a rare scenario and I completely agree that it might not be super important to do it for this feature but because this is new code we are adding, I strongly feel that we do it the right way. It will help in making the API more performant too as you no longer would need to query the complete experiment table each time you update a tag. I understand that this is late in the cycle and would be a little time consuming, therefore I am OK with merging this change provided we have a card to address this in our prioritized backlog. But I would still suggest to do it now if you could spare some time addressing it. Your call! :)

abhishekjain88: Also, in addition to the card, suggest adding a TODO in the code as well.

AndreaSuckro:
@abhishekjain88 : the integration tests should now be passing- the tags were missing in the model copy of the integration test which caused the error. I also rewrote the database structure in a way that we now have one line per experiment in the `experiment_tag` table. I tested locally with unit and integration tests. It would be great if you could have another look at the changes- thanks a lot :)

modules/experiment-objects/src/main/java/com/intuit/wasabi/experimentobjects/NewExperiment.java
```diff
@@ -247,7 +247,7 @@ public void setTags(Set<String> tags) {
         if (null != tags)
             this.tags = new TreeSet<>(tags);
         else
-            this.tags = new TreeSet<>();
+            this.tags = tags;
```
abhishekjain88: @AndreaSuckro Can we make the same change in `PrioritizedExperiment` too?

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/accessor/ExperimentTagAccessor.java
```diff
@@ -31,15 +32,15 @@
...
 
-    @Query("insert into experiment_tag(app_name, tags) values (?,?)")
-    void insertByApp(String appName, Set<String> tags);
+    @Query("insert into experiment_tag(app_name, exp_id, tags) values (?,?,?)")
+    void insert(String appName, UUID experimentId, Set<String> tags);
 
-    @Query("update experiment_tag set tags = tags + ? WHERE app_name = ?")
-    Statement update(Set<String> tags, String appName);
+    @Query("update experiment_tag set tags = tags + ? WHERE app_name = ? and exp_id = ?")
+    Statement update(Set<String> tags, String appName, UUID experimentId);
```
abhishekjain88: @AndreaSuckro Can we rename this method to `addTags()` and the `remove` method to `removeTags()` for better readability.

abhishekjain88: Also, I have a question regarding these methods. I saw that we are completely rewriting the tag list when we add or remove tags. Can we get rid of these methods then?

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/impl/CassandraExperimentRepository.java
```diff
@@ -1253,8 +1249,13 @@ public void createApplication(Application.Name applicationName) {
 
             Map<Application.Name, Set<String>> result = new HashMap<>();
             for (ListenableFuture<Result<ExperimentTagsByApplication>> future : futures) {
-                ExperimentTagsByApplication expTagApplication = future.get().one();
-                result.put(Application.Name.valueOf(expTagApplication.getAppName()), expTagApplication.getTags());
+                List<ExperimentTagsByApplication> expTagApplication = future.get().all();
+
+                Set<String> allTagsForApplication = new TreeSet<>();
```
abhishekjain88: @AndreaSuckro Why use a `TreeSet`? Can we not use a `HashSet()` instead? Do we need any sorted order here? `TreeSet` is expensive in terms of performance (both insert and retrieve)

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/pojo/index/ExperimentTagsByApplication.java
```diff
@@ -37,10 +38,14 @@
     @Column(name = "app_name")
     String appName;
 
+    @PartitionKey(1)
```
abhishekjain88: @AndreaSuckro Pardon my lack of knowledge here! From the migration script looks like exp_id is just the clustering key and not the partition key. Do we declare clustering keys as a partition key in the datastax driver annotation like you've done here?

abhishekjain88: I checked in other places in the code. You might want to probably declare it as a clustering column instead.

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/impl/CassandraExperimentRepository.java
AndreaSuckro: I was using it mainly for the UI, such that @longdogz doesn't need to order it there, but whatever works best :) I also think the performance impact is neglectable due to the small element count.

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/pojo/index/ExperimentTagsByApplication.java
AndreaSuckro: Allright! 👍

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/accessor/ExperimentTagAccessor.java
AndreaSuckro: You are right- I removed them.

AndreaSuckro:
Reran integration and unit tests on the latest changes as well.

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/impl/CassandraExperimentRepository.java
abhishekjain88: @AndreaSuckro That is what I was asking! :) Whether we need the sorting order anywhere, be it the backend or the UI. Now, as you said that is required on the UI side, it makes sense to order it here.

PR #235, opened 2017-05-02
Fix security hole in Feedback dialog

longdogz:
Fixing so we don’t interpret HTML if it is in the feedback as HTML. This was there in order to pass in some bold face for one of the dialogs, so had to change the way the dialog works so I can pass that in that one case, but otherwise, by default, the HTML is safe.  Also fixed so we don’t pass a 0 score to the feedback.


gittify:
:shipit:

PR #240, opened 2017-05-10
Feature/failsafe retry : Exponential backoff based retry mechanism for assignment export

abhishekjain88:


modules/export/src/main/java/com/intuit/wasabi/export/rest/impl/DefaultRestEndPointConfiguration.java
```diff
@@ -93,4 +93,34 @@ public Boolean useProxy() {
     public int getRetries() {
         return Integer.parseInt((String) properties.get("export.rest.retries"));
     }
+
+    /**
```
iizrailevsky: Nice job with Javadoc here! 👍

modules/export/src/main/java/com/intuit/wasabi/export/rest/impl/DefaultRestEndPoint.java
```diff
@@ -17,11 +17,13 @@
 
 import com.google.inject.Inject;
 import com.intuit.wasabi.export.rest.RestEndPoint;
+import net.jodah.failsafe.RetryPolicy;
```
iizrailevsky: Awesome to reuse an existing open source library for this use case!

jcwuzoegiver:
@abhishekjain88 Looks good! Is it being used by wasabi-intuit's Trinity and Omniture?

abhishekjain88:
@jcwuzoegiver Yes, exactly. There is a corresponding PR in wasabi-intuit.

PR #242, opened 2017-05-19
Update README.md

iizrailevsky:
Added doc on packaging without running integration tests, if needed

### DIFF
--- a/README.md
+++ b/README.md
@@ -364,9 +364,22 @@ Code changes can readily be verified by running the growing collection of includ
 
 Wasabi can readily be packaged as installable *rpm* or *deb* distributions and deployed at scale as follows:
 
+> Package by running integration tests 1st:
+
 ```bash
-% ./bin/wasabi.sh package
-% find ./modules -type f \( -name "*.rpm" -or -name "*.deb" \)
+% ./bin/wasabi.sh start package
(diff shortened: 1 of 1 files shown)


abhishekjain88:
Looks good 👍 :shipit: @iizrailevsky

README.md
```diff
@@ -364,9 +364,22 @@ Code changes can readily be verified by running the growing collection of includ
 
 Wasabi can readily be packaged as installable *rpm* or *deb* distributions and deployed at scale as follows:
 
+> Package by running integration tests 1st:
+
+```bash
+% ./bin/wasabi.sh start package
```
bargenilesh: @iizrailevsky Is 'start' needed while packaging (./bin/wasabi.sh start package) ? 
If yes then, does it needed while packing without integration test ?

iizrailevsky: @bargenilesh Yes, we discovered with @abhishekjain88 the 'package' command by default runs integration tests. So it assumes a running docker container with wasabi, started with 'start' command. 

If we're not running integration tests, then there's no need to start containers explicitly.

bargenilesh: @iizrailevsky Got it, thx... Changes looks good... :shipit:

PR #243, opened 2017-05-21
Feature/rapid experiment fix

abhishekjain88:
This change fixes the over assignment of users in rapid experiments beyond the user cap - due to stale experiment state in experiment metadata cache.

### DIFF
--- a/modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
+++ b/modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
@@ -37,11 +37,13 @@ import com.intuit.wasabi.assignmentobjects.RuleCache;
 import com.intuit.wasabi.assignmentobjects.SegmentationProfile;
 import com.intuit.wasabi.assignmentobjects.User;
 import com.intuit.wasabi.eventlog.EventLog;
+import com.intuit.wasabi.eventlog.events.ExperimentChangeEvent;
 import com.intuit.wasabi.exceptions.AssignmentExistsException;
 import com.intuit.wasabi.exceptions.BucketDistributionNotFetchableException;
 import com.intuit.wasabi.exceptions.BucketNotFoundException;
 import com.intuit.wasabi.exceptions.ExperimentNotFoundException;
 import com.intuit.wasabi.exceptions.InvalidAssignmentStateException;
+import com.intuit.wasabi.experiment.Experiments;
 import com.intuit.wasabi.experiment.Pages;
 import com.intuit.wasabi.experiment.Priorities;
 import com.intuit.wasabi.experimentobjects.Application;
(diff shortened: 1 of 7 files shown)


gittify:
Can be merged with appropriate tests and also needs performance tests :shipit:

modules/repository-datastax/src/test/java/com/intuit/wasabi/repository/cassandra/impl/AssignmentCountEnvelopeTest.java
```diff
@@ -53,7 +53,7 @@ public void testErrorInUpdateBucket() {
         env.run();
         // so since the two parts a separate code fragments the fail of the first
         // should not result in a fail of the other
-        verify(exp, times(1)).getIsRapidExperiment();
+        verify(exp, times(2)).getIsRapidExperiment();
```
iizrailevsky: Nice way to verify it twice! 👍

modules/repository-datastax/src/test/java/com/intuit/wasabi/repository/cassandra/impl/AssignmentCountEnvelopeTest.java
```diff
@@ -77,7 +77,7 @@ public void testRapidExperimentationLimitNotReached() {
 //        doReturn(42).when(exp).getUserCap();
         when(exp.getIsRapidExperiment()).thenReturn(true);
         when(exp.getUserCap()).thenReturn(42);
-        when(ar.getBucketAssignmentCount(exp).getTotalUsers().getBucketAssignments()).thenReturn(41l);
+        when(ar.getBucketAssignmentCount(exp).getTotalUsers().getBucketAssignments()).thenReturn(40l);
```
iizrailevsky: Why is it 40, instead of 41, in this case?

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -609,6 +610,15 @@ protected Assignment getAssignment(User.ID userID, Application.Name applicationN
             if (createAssignment) {
                 if (experiment.getState() == Experiment.State.PAUSED) {
                     return nullAssignment(userID, applicationName, experimentID, Assignment.Status.EXPERIMENT_PAUSED);
+                } else if (experiment.getIsRapidExperiment() != null && experiment.getIsRapidExperiment()) {
+                    //Get the latest state from the DB for rapid experiments if they are not in stopped state.
+                    Experiment rapidExperiment = experimentUtil.getExperiment(experiment.getID());
+                    Experiment.Label rapidExperimentLabel = null != rapidExperiment ? rapidExperiment.getLabel(): null;
```
iizrailevsky: Mind wrapping the conditional in (...) to make the assignment more readable? E.g.:

Experiment.Label rapidExperimentLabel = (rapidExperiment != null ? rapidExperiment.getLabel() : null);

modules/repository-datastax/src/test/java/com/intuit/wasabi/repository/cassandra/impl/AssignmentCountEnvelopeTest.java
abhishekjain88: @iizrailevsky Earlier the check for cap used to happen after updating the bucket count. With the latest code changes, it will happen before the update. So, there is a difference of 1 in the count which needs to be modified in the test. We are verifying "limit not reached here". If the limit was "42" and this method returns "41" already, and the check happens for "userCap - 1" before updating the count, it will say "limit reached". Hope it makes sense.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
abhishekjain88: @iizrailevsky Done!

modules/repository-datastax/src/test/java/com/intuit/wasabi/repository/cassandra/impl/AssignmentCountEnvelopeTest.java
iizrailevsky: Got it - thanks! 👍

PR #246, opened 2017-05-22
Adds option for console only logging for run script

mgauthier:


mgauthier:
@abhishekjain88 I see what you are saying, happy to give that a shot. That being said, logging to console only should be a pretty common use case so do you think there is value in keeping the changes as is and adding the ability to specify a custom logback file location, or would you prefer that we replace this change completely with what you are suggesting?

abhishekjain88:
@mgauthier I would still prefer replacing the change if possible. Would look forward to a PR from your side! :)

PR #247, opened 2017-05-24
Modifies ui build to allow for env vars to be inserted for api and ui…

mgauthier:
… config


bin/wasabi.sh
```diff
@@ -302,7 +302,7 @@ package() {
    done)
   (cd modules/ui; \
     mkdir -p target; \
-    for f in app node_modules bower.json Gruntfile.js constants.json karma.conf.js karma-e2e.conf.js package.json test .bowerrc; do \
+    for f in build/constants.json app node_modules bower.json Gruntfile.js default_constants.json karma.conf.js karma-e2e.conf.js package.json test .bowerrc; do \
```
longdogz: I don't think this makes sense?  That is, the "build/constants.json".  This is created by the grunt build (specifically, the config:dist step) and that is done *after* this step in wasabi.sh.  So the build/constants.json doesn't exist, yet, does it?

mgauthier: Oops I believe you're right, removed

modules/ui/README.md
```diff
@@ -38,7 +38,7 @@ Install Yo, Grunt, Bower, and Compass
 If you want to test how the UI runs from the combined and minified files (which you would generally use
 when you go to Production), you will need to do the following (after you have done the build steps above):
 
-Edit constants.json with the following change to the apiHostBaseUrlValue value, since you would be running the
+Edit default_constants.json with the following change to the apiHostBaseUrlValue value, since you would be running the
```
longdogz: This is fine, but actually, I think I need to update this part of the README to reflect the change with the default value of "DEFAULT", which causes the UI to use the URL it was served from as the default backend URL.

mgauthier: Ok, are you referring to the root level README in the project? Do I need to make another change somewhere else?

longdogz: Actually, no.  I was referring to this README.  But I take it back.  This comment is still relevant as when you are serving the UI on localhost:9000, but the backend is localhost:8080, then you DO need to do this edit.  So, never mind.

PR #248, opened 2017-05-24
Feature/reimplement rapid experiment fix

abhishekjain88:
Refer PR: https://github.com/intuit/wasabi/pull/243 

@bargenilesh suggested that we might want to just query the bucket assignment count once to decide on whether to do the assignment or not for rapid experiments and thereby, avoid the additional call to the database to fetch the experiment state in the core assignment flow.

Since, this was a major change from the old code flow, therefore creating a new PR/branch to address the same.


iizrailevsky:
Looks good! 👍 :shipit:

modules/experiment/src/main/java/com/intuit/wasabi/experiment/impl/ExperimentsImpl.java
```diff
@@ -578,4 +578,28 @@ public Experiment updateExperiment(final Experiment.ID experimentID, final Exper
 
         return experiment;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void updateExperimentState(final Experiment experiment, final Experiment.State state) {
```
bargenilesh: Nice flow to make Cassandra & MySQL single transaction... 👍

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/impl/CassandraAssignmentsRepository.java
```diff
@@ -482,9 +482,9 @@ public void assignUsersInBatch(List<Pair<Experiment, Assignment>> assignments, D
      */
     private void incrementCounts(List<Pair<Experiment, Assignment>> assignments, Date date) {
         boolean countUp = true;
-        assignments.forEach(pair -> {
-            assignmentsCountExecutor.execute(new AssignmentCountEnvelope(this, experimentRepository, dbRepository, pair.getLeft(), pair.getRight(), countUp, eventLog, date, assignUserToExport, assignBucketCount));
-        });
+        assignments.forEach(pair -> assignmentsCountExecutor.execute(new AssignmentCountEnvelope(
```
bargenilesh: Is there any logical change in code here or just formatting? You may want to check the code style formatter...

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -660,6 +662,27 @@ protected Assignment getAssignment(User.ID userID, Application.Name applicationN
     }
 
     /**
+     * If the experiment is a rapid experiment, it checks whether the user cap has already been achieved.
+     * If yes, then it sets the experiment state to paused, refreshes the experiment metadata cache
+     * and returns false i.e. assignment is not allowed in that case.
+     */
+    private boolean checkUserCapForRapidExperiment(Experiment experiment) {
+        if (experiment.getIsRapidExperiment() != null && experiment.getIsRapidExperiment()) {
```
bargenilesh: Minor: Just for consistency, you can use Object.isNull() method...

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/impl/CassandraAssignmentsRepository.java
abhishekjain88: @bargenilesh There was a change earlier. I reverted it in the latest change but as I touched this class, the original code was going way above the monitor width, therefore I split it into 2 lines for better readability. The code formatter does not complain even after the split, so should be fine, I guess! :)

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -660,6 +662,27 @@ protected Assignment getAssignment(User.ID userID, Application.Name applicationN
     }
 
     /**
+     * If the experiment is a rapid experiment, it checks whether the user cap has already been achieved.
+     * If yes, then it sets the experiment state to paused, refreshes the experiment metadata cache
+     * and returns false i.e. assignment is not allowed in that case.
+     */
+    private boolean checkUserCapForRapidExperiment(Experiment experiment) {
```
bargenilesh: Major: This check needs to be done in putAssignment() flow as well...

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
abhishekjain88: @bargenilesh Can we just agree on using the vanilla null check everywhere. As we had discussed in earlier code reviews as well, the Object.isNull() method was introduced explicitly for the streams API in Java 8.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
abhishekjain88: @bargenilesh AFAIK, the PUT assignment flow is for admin use to update assignments only. Do we want to really enforce this check here?

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -607,7 +609,7 @@ protected Assignment getAssignment(User.ID userID, Application.Name applicationN
         Assignment assignment = getAssignment(experimentID, userID, context, userAssignments, bucketList);
         if (assignment == null || assignment.isBucketEmpty()) {
             if (createAssignment) {
-                if (experiment.getState() == Experiment.State.PAUSED) {
+                if (experiment.getState() == Experiment.State.PAUSED || !checkUserCapForRapidExperiment(experiment)) {
```
bargenilesh: Regarding, Place where we are checking the user assignment cap: it would be great if experiment caps are retrieved from database asynchronously and then checked here synchronously in a loop one by one in a priority order...

iizrailevsky: @abhishekjain88: just a nit: should the method name be more descriptive, as opposed to just checking: "checkUserCapForRapidExperiment(...)"? As in, "isUserCapForRapidExperimentExceeded(...)"?

abhishekjain88: @iizrailevsky Done.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
abhishekjain88: @bargenilesh Done for the PUT API as well.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
abhishekjain88: @bargenilesh Implemented the fetch of counts asynchronously.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -607,7 +632,8 @@ protected Assignment getAssignment(User.ID userID, Application.Name applicationN
         Assignment assignment = getAssignment(experimentID, userID, context, userAssignments, bucketList);
         if (assignment == null || assignment.isBucketEmpty()) {
             if (createAssignment) {
-                if (experiment.getState() == Experiment.State.PAUSED) {
+                if (experiment.getState() == Experiment.State.PAUSED ||
+                        !isUserCapForRapidExperimentReached(experiment, rapidExperimentToAssignmentCounts)) {
```
bargenilesh: @abhishekjain88 As the name of this method changed to isUserCapForRapidExperimentReached(), should it now return TRUE when user cap is reached else FALSE.

abhishekjain88: @bargenilesh Thats a good point. Should I instead rename it to "isUserCapForRapidExperimentNotReached()"? Given that it has already undergone functional testing, I don't want to touch the core logic. What do you think? Let me know.

bargenilesh: @abhishekjain88  isAssignmentValidForRapidExperiment() sounds easier to understand...

abhishekjain88: cool @bargenilesh !

bargenilesh:
@abhishekjain88 Awesome job 👍  and :shipit:  provided all functional & unit tests passes...

PR #249, opened 2017-05-25
Feature/vagrant integration test on jenkins

bargenilesh:
Following changes are done to make sure functional test run successfully on Vagrant VM and on Jenkins:
 - Increase Cassandra connection & request timeout values
 - Increased resources (memory & cpu) for vagrant VM

Also, had to update JDK location...


jcwuzoegiver:
Looks good :shipit:

PR #250, opened 2017-05-26
Feature/qe automation rapidexp fix

sudheerkethireddy:
This PR has the automation that checks for rapid experiments for various scenarios. The scanerios covered are
1. single experiment with rapid exp enabled 
2. Single experiment that has been changed to rapid exp while doing assignments
3. Batch assignment consisting of all rapid experiments
4. Batch assignment consisting of a mixture of rapid experiments and normal experiments.
5. Test putassignment call for a simple experiment


modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,328 @@
...
+            // lets assert the status and response code and also the state of the experiment
+            assertReturnCode(response, HttpStatus.SC_OK);
+            Assert.assertEquals(assignment.status, "NEW_ASSIGNMENT");
+            Assert.assertEquals(exp.state, Constants.EXPERIMENT_STATE_RUNNING);
+        }
+
+        exp = exp.setIsRapidExperiment(true).setUserCap(3);
+        exp = putExperiment(exp);
+
+        clearAssignmentsMetadataCache();
```
abhishekjain88: @sudheerkethireddy I don't think this will work when we run integration tests against a wasabi cluster instead of a single node. Can we put a note here or create a test profile where we disable this test when running against a cluster?

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,328 @@
...
+     * This test case tests for the put assignment API call that forces assignment call on particular bucket for a rapid
+     * experiment
+     */
+    @Test
+    public void testPutAssignmentExperimentToRapidExperiment() {
+        Experiment exp = getExperiment(expList.get(2));
+        List<Bucket> bucketList = getBuckets(exp);
+        Assignment assignment = AssignmentFactory.createAssignment().setAssignment(bucketList.get(0).label)
+                .setExperimentLabel(exp.label).setOverwrite(true);
+        for (int i = 1; i <= 5; i++) {
```
abhishekjain88: @sudheerkethireddy Can we please replace the hardcoded `5` with the constant everywhere in the tests?

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,328 @@
...
+            assertReturnCode(response, HttpStatus.SC_OK);
+            Assert.assertEquals(assignment.status, "NEW_ASSIGNMENT");
+            Assert.assertEquals(exp.state, Constants.EXPERIMENT_STATE_RUNNING);
+        }
+
+        // since the experiment crossed the maxusers the experiment should change to PUASED
+        assignment = putAssignment(exp, assignment, new User("user6"), "PROD");
+        assertReturnCode(response, HttpStatus.SC_OK);
+
+        // Assignment assignment2 = postAssignment(exp, new User("user7"), "PROD");
```
abhishekjain88: @sudheerkethireddy Please remove commented out code.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,328 @@
...
+        for (int i = 1; i <= 5; i++) {
+            // lets do an assignment for a user
+            assignment = putAssignment(exp, assignment, new User("user" + i));
+            // lets assert the status and response code and also the state of the experiment
+            assertReturnCode(response, HttpStatus.SC_OK);
+            Assert.assertEquals(assignment.status, "NEW_ASSIGNMENT");
+            Assert.assertEquals(exp.state, Constants.EXPERIMENT_STATE_RUNNING);
+        }
+
+        // since the experiment crossed the maxusers the experiment should change to PUASED
```
abhishekjain88: @sudheerkethireddy Typo: Change to PAUSED.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,328 @@
...
+        Assert.assertEquals(exp.state, Constants.EXPERIMENT_STATE_PAUSED);
+        expList.set(1, rapidExperiment2);
+    }
+
+    /**
+     * This test case tests for the put assignment API call that forces assignment call on particular bucket for a rapid
+     * experiment
+     */
+    @Test
+    public void testPutAssignmentExperimentToRapidExperiment() {
```
abhishekjain88: @sudheerkethireddy Please rename to `testPutAssignmentToRapidExperiment`

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,328 @@
...
+
+        // since the experiment crossed the maxusers the experiment should change to PUASED
+        assignment = putAssignment(exp, assignment, new User("user6"), "PROD");
+        assertReturnCode(response, HttpStatus.SC_OK);
+
+        // Assignment assignment2 = postAssignment(exp, new User("user7"), "PROD");
+        // assertReturnCode(response, HttpStatus.SC_OK);
+        // Assert.assertEquals(assignment2.status, "EXPERIMENT_PAUSED");
+
+        assignment = putAssignment(exp, assignment, new User("user8"), "PROD");
```
abhishekjain88: @sudheerkethireddy Can we please put these multiple `putAssignment` calls in a for loop and check for the experiment state in each iteration.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,328 @@
...
+            for (Assignment assignment : assignments) {
+                Assert.assertEquals(assignment.status, "NEW_ASSIGNMENT");
+
+            }
+        }
+
+        // if we do one more assignment the rapid experiments should change to PAUSED/STOPPED state and the remaining
+        // should be in running state
+        List<Assignment> assignments = postAssignments(application, partialpage, new User("user1000"));
+        for (int i = 1; i <= RAPID_EXP_MAX_USERS; i++) {
```
abhishekjain88: @sudheerkethireddy This loop should run on the number of experiments and not the user cap of rapid experiments. Also, as I said earlier, it might be a better idea to iterate using `for (Experiment experiment: batchExperiments)` than this loop.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,328 @@
...
+            Experiment exp = getExperiment(batchExperiments.get(i - 1));
+            if (i <= NUMBER_OF_RAPID_EXPERIMENTS_IN_BATCH) {
```
abhishekjain88: @sudheerkethireddy Instead of basing this check on the position of the experiment in the list, you could check whether the experiment is a rapid experiment or not and based on that, do the asserts.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,328 @@
...
+     * This test case covers a scenario where we do batch assignment for a group of experiments that has all rapid
+     * experiments. It validates the behavior when we do bulk assignment of a user to that batch
+     */
+    @Test
+    public void testAssignmentForBatchWithAllRapidExperiments() {
+
+        page = new Page(BATCH_PAGE, true);
+        List<Experiment> expList = new ArrayList<Experiment>();
+        // lets create experiments for the batch
+        for (int i = 1; i <= NUMBER_OF_EXPERIMENTS_PER_BATCH; i++) {
```
abhishekjain88: @sudheerkethireddy A lot of this code is duplicate to the code in the previous test. You might want to put the common code in a private method and modify the "rapid experiment" property selectively on the list of experiments the common code returns.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,328 @@
...
+            for (Assignment assignment : assignments) {
+                Assert.assertEquals(assignment.status, "NEW_ASSIGNMENT");
+
+            }
+        }
+
+        // if we do one more assignment the rapid experiments should change to PAUSED/STOPPED state and the remaining
+        // should be in running state
+        List<Assignment> assignments = postAssignments(application, page, new User("user100"));
+        for (int i = 1; i <= RAPID_EXP_MAX_USERS; i++) {
```
abhishekjain88: @sudheerkethireddy This loop should run on the number of experiments. Also, consider replacing it with `for (Experiment experiment: expList)` as I mentioned earlier.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,328 @@
+package com.intuit.wasabi.tests.service;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+
+import org.apache.http.HttpStatus;
+import org.testng.Assert;
+import org.testng.annotations.AfterClass;
```
sudheerkethireddy: done (y)

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,328 @@
...
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *******************************************************************************/
+public class RapidExperimentationTest extends TestBase {
+    /**
+     * This test class contains functional tests that check the rapid experimentation functionality
+     * 
+     * Created on 05/23/2017
+     */
+    private static final int RAPID_EXP_MAX_USERS = 5; // max cap on number of users
```
sudheerkethireddy: i think this would be better as i dont want to overload the config file and making it confusing..since this constant is only for rapid experimentation test...i feel its better we use the constants

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,328 @@
...
+    private static final int NUMBER_OF_EXPERIMENTS_PER_BATCH = 5;
+    private static final int NUMBER_OF_RAPID_EXPERIMENTS_IN_BATCH = 2; // number of rapid experiments in a batch that
```
sudheerkethireddy: The justification is that I  think this would be better as i dont want to overload the config file and making it confusing..since this constant is only for rapid experimentation test...i feel its better we use the constants

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
sudheerkethireddy: good suggestion @abhishekjain88  (y)

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
sudheerkethireddy: good one...i missed out on that my bad (y) @abhishekjain88

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,328 @@
...
+                                                                       // has combination of normal and rapid
+                                                                       // experiments
+
+    private static final String BATCH_PAGE = "rapidexperiment_page";
+    private static final String PARTIAL_BATCH_PAGE = "rapidexperiment_partialpage";
+    Experiment rapidExperiment1 = null;
+    Experiment rapidExperiment2 = null;
+    Experiment rapidExperiment3 = null;
+    Page page = null;
+    List<Experiment> expList = new ArrayList<Experiment>();
```
sudheerkethireddy: ack: @abhishekjain88

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,328 @@
...
+
+    @BeforeClass
+    public void setUp() {
+
+        application = new Application("rapidExperimentApplication");
+
+        experimentsListOfRapidApplication = getApplicationExperiments(application);
+        pauseTerminateAndDeleteExperiments(experimentsListOfRapidApplication);
+
+        rapidExperiment1 = ExperimentFactory.createExperiment("_1" + UUID.randomUUID().toString(), -1)
```
sudheerkethireddy: @abhishekjain88 i meant _1 only but anyhow @abhishekjain88 1_ is also good suggestion :)

sudheerkethireddy: @abhishekjain88  i wanted it 1_ only but wasabi doesnt allow experiment labels to start with digit hence i had to use _1

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,328 @@
...
+                .setApplication(application).setIsRapidExperiment(true).setUserCap(RAPID_EXP_MAX_USERS);
+        rapidExperiment2 = ExperimentFactory.createExperiment("_2" + UUID.randomUUID().toString(), -1)
```
sudheerkethireddy: Good catch changed it to experimentChangedToRapidExperiment @abhishekjain88

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,328 @@
...
+                .setApplication(application).setIsRapidExperiment(true).setUserCap(RAPID_EXP_MAX_USERS);
+
+        rapidExperiment1 = postExperiment(rapidExperiment1);
+        rapidExperiment2 = postExperiment(rapidExperiment2);
+        rapidExperiment3 = postExperiment(rapidExperiment3);
+        expList.add(rapidExperiment1);
+        expList.add(rapidExperiment2);
+        expList.add(rapidExperiment3);
+        for (int i = 1; i <= 3; i++) {
+            Experiment exp = expList.get(i - 1);
```
sudheerkethireddy: thought of it....but somehow forgot....changed now

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,328 @@
...
+            List<Bucket> bucketList = BucketFactory.createBuckets(exp, 3);
```
sudheerkethireddy: the enhanced for loop will take care of this

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
sudheerkethireddy: The reason i havent done is so that i can reuse the same looping value to get the assignment as well hence decreasing number of lines of code

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
abhishekjain88: @sudheerkethireddy I agree but using constants to specify these values involve modifying the code to change these values. Making it configurable is a nice to have where you just have to make changes in config files. Also, you could have multiple property files. For example, you could use a separate property file for rapid experiment test and let the global config file remain as is. We probably need to start doing it for existing tests too where they could each have their own config file. 

I would still suggest loading these from a property file as this is a new test that we are adding and it will lead the way for other tests. Should be simple to load these values from a property file: https://www.mkyong.com/java/java-properties-file-examples/ 

The autumn library: https://github.com/intuit/Autumn also abstracts property loading as we use in Wasabi code as well. For example, refer to the `configure()` method in: https://github.com/intuit/wasabi/blob/develop/modules/assignment/src/main/java/com/intuit/wasabi/assignment/AssignmentsModule.java. There is a usage around `getProperty()` from autumn.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
abhishekjain88: @sudheerkethireddy Please refer to my comment above!

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
abhishekjain88: @sudheerkethireddy Can we then change it to `_1_` ? It separates the uuid from the number which is attached and makes it easy to distinguish.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
abhishekjain88: @sudheerkethireddy I don't think that is the best idea. You should check for the experiment state and the assignment status separately as they are disjoint with each other. The `assignments` list and the `expList` are 2 different list data structures and may vary in their sizes and position of elements. As long as you have the `RAPID_EXP_MAX_USERS` same as `NUMBER_OF_EXPERIMENTS_PER_BATCH`, this will continue to work but as soon as these 2 values get changed in future and diverge, you will end up hitting `ArrayIndexOutOfBoundsException`. 

For example, say `RAPID_EXP_MAX_USERS` gets modified to 20 and the number of experiments stays at 5 then you will end up trying to retrieve a value at index 6-19 on expList which will throw an `ArrayIndexOutOfBoundsException`.

We can sync up offline if you don't understand.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,328 @@
...
+            postBuckets(bucketList);
+
+            exp.state = Constants.EXPERIMENT_STATE_RUNNING;
+            exp = putExperiment(exp);
```
sudheerkethireddy: ack: @abhishekjain88

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
sudheerkethireddy: ok sure will do @abhishekjain88

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,328 @@
...
+
+        }
+
+    }
+
+    @Test
+    public void testRapidExperiment() {
+
+        Experiment exp = getExperiment(expList.get(0));
+        for (int i = 1; i <= 5; i++) {
```
sudheerkethireddy: done....

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,328 @@
...
+            exp = putExperiment(exp);
+
+        }
+
+    }
+
+    @Test
+    public void testRapidExperiment() {
+
+        Experiment exp = getExperiment(expList.get(0));
```
sudheerkethireddy: Good suggestion... i will make each test more atomic....so that experiment creation will be part of the test...

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,328 @@
...
+        Assert.assertEquals(assignment.status, "EXPERIMENT_PAUSED");
+        assignment = postAssignment(exp, new User("user11"), "PROD");
+        assertReturnCode(response, HttpStatus.SC_OK);
+        Assert.assertEquals(assignment.status, "EXPERIMENT_PAUSED");
+        assignment = postAssignment(exp, new User("user12"), "PROD");
+        assertReturnCode(response, HttpStatus.SC_OK);
+        Assert.assertEquals(assignment.status, "EXPERIMENT_PAUSED");
+        exp = getExperiment(exp);
+        Assert.assertEquals(exp.state, Constants.EXPERIMENT_STATE_PAUSED);
+        expList.set(0, rapidExperiment1);
```
sudheerkethireddy: Now this code is refactored we no longer need this as each test is made atomic...

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
sudheerkethireddy: Ack..implemneted it...thanks for the suggestion..

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
sudheerkethireddy: done...implemented it...

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
sudheerkethireddy: ack  @abhishekjain88  in the new code checking for rapid experiment rather than doing by index..

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
sudheerkethireddy: done...

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,315 @@
...
+    public void setUp() {
+
+        // lets terminate all the experiments that are created by the previous run of test
+        application = new Application(rapidExperimentTestApplication);
+        experimentsListOfRapidApplication = getApplicationExperiments(application);
+        pauseTerminateAndDeleteExperiments(experimentsListOfRapidApplication);
+
+    }
+
+    @Test
```
bargenilesh: Adding test description here would be great...

sudheerkethireddy: sure ack @bargenilesh will do it..

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,315 @@
...
+
+            exp = createBucketsAndRunExperiment(exp, NUMBER_OF_BUCKETS_PER_EXPERIMENT);
+            batchExperiments.add(exp);
+
+        }
+        assignPageToExperimentsList(batchExperiments, partialpage);
+
+        // the assignment will happen until the max users cap is reached for all experiments in batch
+        for (int i = 1; i <= RAPID_EXP_MAX_USERS; i++) {
+            List<Assignment> assignments = postAssignments(application, partialpage, new User("user" + i));
```
bargenilesh: 👍  for use of page based assignments as well...

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,315 @@
...
+            assertAssignmentResponseStatus(assignments, "NEW_ASSIGNMENT");
+        }
+
+        // if we do one more assignment the rapid experiments should change to PAUSED/STOPPED state and the remaining
+        // should be in running state
+        List<Assignment> assignments = postAssignments(application, partialpage, new User("user1000"));
+
```
bargenilesh: Can you also assert here on the statuses of returned assignments for those rapid experiments?

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
```diff
@@ -0,0 +1,315 @@
...
+            Experiment exp = getExperiment(experiment);
+            Assert.assertEquals(exp.state, Constants.EXPERIMENT_STATE_PAUSED);
+        }
+
+        // assert that the response of the post assignments are EXPERIMENT_PAUSED
+        assertAssignmentResponseStatus(assignments, "EXPERIMENT_PAUSED");
+
+    }
+
+    /**
```
bargenilesh: Do you think, it would be useful for all below 4 methods to move to TestBase class so that they can used used across other test classes when needed?

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
sudheerkethireddy: done...

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
sudheerkethireddy: the postAssignments intrinsically asserts the status of response code to HttpStatus.SC_CREATED

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
sudheerkethireddy: thanks

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/RapidExperimentationTest.java
sudheerkethireddy: yeah thats a good idea will move the util functions to TestBase

bargenilesh:
@sudheerkethireddy changes looks good, Plz sync-up with @abhishekjain88 about his comment - I feel it is a good idea but involves additional efforts which needs to be planned... :shipit: provided other dependent changes and successful completion of unit & integration tests...

PR #254, opened 2017-06-07
Changes to bring the Tools menu on top of opened Running dialogs.

longdogz:
NOTE: This also includes a .gitinclude change that was needed because of the grunt build changes that generate a build/constants.json file, instead of that being the name of the file in modules/ui (which was changed to default_constants.json .


sudheerkethireddy:
looks good :shipit:

PR #261, opened 2017-07-16
adding handling for empty tag list and unit test

AndreaSuckro:
These changes fix the NP exception that is shown on the UI side when all Experiments are retrieved, but some applications have no associated tags stored with them.

### DIFF
--- a/modules/functional-test/src/main/java/com/intuit/wasabi/tests/library/TestBase.java
+++ b/modules/functional-test/src/main/java/com/intuit/wasabi/tests/library/TestBase.java
@@ -3701,6 +3701,19 @@ public class TestBase extends ServiceTestBase {
     }
 
     /**
+     * Sends a GET request to retrieve all experiment tags for any application. The response must contain HTTP
+     * {@code HttpStatus.SC_OK}.
+     *
+     * @return an experiment
+     */
+    public String getExperimentTags() {
+        String uri = "applications/tags";
+        response = apiServerConnector.doGet(uri);
+        assertReturnCode(response, HttpStatus.SC_OK);
+        return response.jsonPath().prettify();
+    }
+
+    /**
      * Sends a POST request to receive statistics for an experiment. The response must contain {@link HttpStatus#SC_OK}.
      *
      * @param experiment the experiment
--- a/modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/ApplicationTagListIntegrationTest.java
+++ b/modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/ApplicationTagListIntegrationTest.java
@@ -1,0 +1,73 @@
+/*******************************************************************************
+ * Copyright 2017 Intuit
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+package com.intuit.wasabi.tests.service;
+
+import com.intuit.wasabi.tests.library.TestBase;
+import com.intuit.wasabi.tests.library.util.serialstrategies.DefaultNameExclusionStrategy;
+import com.intuit.wasabi.tests.model.Experiment;
+import com.intuit.wasabi.tests.model.factory.ExperimentFactory;
+import org.apache.http.HttpStatus;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+import java.util.UUID;
+
+import static com.intuit.wasabi.tests.library.util.Constants.EXPERIMENT_STATE_DELETED;
+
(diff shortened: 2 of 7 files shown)


bargenilesh:
@AndreaSuckro Changes looks good 👍  Provided unit, integration tests & UI works :shipit:

modules/functional-test/src/main/java/com/intuit/wasabi/tests/library/TestBase.java
```diff
@@ -3701,6 +3701,19 @@ public ExperimentCumulativeCounts getExperimentCumulativeCounts(Experiment exper
...
+     * @return an experiment
+     */
+    public String getExperimentTags() {
+        String uri = "applications/tags";
+        response = apiServerConnector.doGet(uri);
+        assertReturnCode(response, HttpStatus.SC_OK);
+        return response.jsonPath().prettify();
+    }
+
+    /**
```
sudheerkethireddy: Good work on extending the testbase and adding this method

sudheerkethireddy:
Looks good @AndreaSuckro. :shipit:

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/ApplicationTagListIntegrationTest.java
```diff
@@ -0,0 +1,73 @@
...
+public class ApplicationTagListIntegrationTest extends TestBase {
+
+    private Experiment experiment;
+    private String appName = "tagListTest-" + UUID.randomUUID();
+
+    /**
+     * Set up one experiment that has no tags.
+     */
+    @BeforeClass
+    public void setUp() {
```
abhishekjain88: @AndreaSuckro I understand that this creates an experiment without tags? In my opinion, we should have tests for both scenarios i.e. Happy path (experiment with tags) and negative path (experiment without tags).

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/ApplicationTagListIntegrationTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+    }
+
+
+    @Test(dependsOnGroups = {"ping"})
+    public void t_RetrieveTagList() {
+        //creates an experiment for that app
+        this.experiment = postExperiment(experiment);
+        //retrieve response (method checks for status code OK)
+        String response = getExperimentTags();
```
abhishekjain88: @AndreaSuckro In the case where you would add a test for experiment with tags, please also validate at the end of the test that you are getting back the exact same list of tags that you had posted while setting up the test.

modules/functional-test/src/main/java/com/intuit/wasabi/tests/service/ApplicationTagListIntegrationTest.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @AfterClass
+    public void terminateAndDelete() {
```
abhishekjain88: @AndreaSuckro Don't we have a standard utility method across all tests which takes care of terminating and deleting an experiment? Can you please check on that. If not, let's add that utility method and use that from all tests to avoid duplication of code and we having to update code at several places in case of specifications changing in the future. cc @sudheerkethireddy

modules/repository-datastax/src/test/java/com/intuit/wasabi/repository/cassandra/impl/CassandraExperimentRepositoryTest.java
```diff
@@ -722,4 +723,31 @@ public void testGetTagList() throws ExecutionException, InterruptedException {
...
+        //------ Input --------
+        Application.Name app01 = Application.Name.valueOf("app01");
+        List<Application.Name> appNames = Arrays.asList(app01);
+
+        List<ExperimentTagsByApplication> exp1Tags = Collections.emptyList();
+
+        //------ Mocking interacting calls
+        ListenableFuture<Result<ExperimentTagsByApplication>> dbResultFuture1 = mock(ListenableFuture.class);
+
+        Result<ExperimentTagsByApplication> dbResult1 = mock(Result.class);
```
abhishekjain88: @AndreaSuckro Let us just keep it as `dbResult` or rather `dbResultMock` instead of `dbResult1`.

modules/repository-datastax/src/test/java/com/intuit/wasabi/repository/cassandra/impl/CassandraExperimentRepositoryTest.java
```diff
@@ -722,4 +723,31 @@ public void testGetTagList() throws ExecutionException, InterruptedException {
         assertArrayEquals(exp2.toArray(), callResult.get(app02).toArray());
     }
 
+
+    @Test
+    public void testGetEmptyTagList() throws ExecutionException, InterruptedException {
+        //------ Input --------
+        Application.Name app01 = Application.Name.valueOf("app01");
```
abhishekjain88: @AndreaSuckro There is only 1 app, 1 experiment and 1 db result in this test. Let's do away with the number identifier at the end of each of those variables.

Also, I took a look at the previous test as well (which was checked in earlier - apologies for not suggesting it then)  - Let us try to name variables as meaningfully as we can instead of using numbers at the end of variables. For example `dbResult2` could be changed to `dbResultforapp2` etc. etc.

abhishekjain88: It improves readability of the code in general and it becomes easier for anybody else to understand later in case any modifications or debugging needs to be done later.

PR #262, opened 2017-07-17
Feature/state remodel

AndreaSuckro:
If an experiment reaches the enddate it is currently not possible to further extend the duration. The experiment did move to a semi-stopped state where the `Assignment.Status` returned that the experiment expired. 

This branch changes the behavior for an expiring experiment in the sense that it is moved to PAUSED where otherwise the `Assignment.Status.EXPERIMENT_EXPIRED` was returned. The following Assignments get the `EXPERIMENT_PAUSED` state returned. The old Assignment.Status is kept for the historical records.

### DIFF
--- a/modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
+++ b/modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
@@ -59,7 +59,6 @@ import java.util.Map;
 import static com.google.common.collect.Maps.newHashMap;
 import static com.intuit.wasabi.api.APISwaggerResource.DEFAULT_LABELLIST;
 import static com.intuit.wasabi.api.APISwaggerResource.EXAMPLE_AUTHORIZATION_HEADER;
-import static com.intuit.wasabi.assignmentobjects.Assignment.Status.EXPERIMENT_EXPIRED;
 import static java.lang.Boolean.FALSE;
 import static java.util.Objects.isNull;
 import static java.util.Objects.nonNull;
@@ -734,7 +733,7 @@ public class AssignmentsResource {
         }
 
         // Only include `assignment` property if there is a definitive assignment, either to a bucket or not
-        if (assignment.getStatus() != EXPERIMENT_EXPIRED) {
+        if (assignment.getStatus().isDefinitiveAssignment()) {
             response.put("assignment",
                     nonNull(assignment.getBucketLabel()) ? assignment.getBucketLabel().toString() : null);
 
--- a/modules/assignment-objects/src/main/java/com/intuit/wasabi/assignmentobjects/Assignment.java
+++ b/modules/assignment-objects/src/main/java/com/intuit/wasabi/assignmentobjects/Assignment.java
@@ -284,7 +284,6 @@ public class Assignment {
         EXPERIMENT_IN_DRAFT_STATE(false),
         EXPERIMENT_PAUSED(false),
         NO_PROFILE_MATCH(false),
-        EXPERIMENT_EXPIRED(false),
         ASSIGNMENT_FAILED(false),
 
         EXISTING_ASSIGNMENT(true),
--- a/modules/assignment-objects/src/main/java/com/intuit/wasabi/assignmentobjects/Assignment.java
+++ b/modules/assignment-objects/src/main/java/com/intuit/wasabi/assignmentobjects/Assignment.java
@@ -285,7 +285,8 @@ public class Assignment {
         EXPERIMENT_PAUSED(false),
         NO_PROFILE_MATCH(false),
         // EXPERIMENT_EXPIRED is no longer used since the experiment is moved to PAUSED upon expiring.
-        // The status is kept for the historical data.
+        // The status is kept for the historical data in the downstream systems, but is not needed
+        // by Wasabi's application storage (Cassandra & Mysql)
         EXPERIMENT_EXPIRED(false),
         ASSIGNMENT_FAILED(false),
 
--- a/modules/assignment-objects/src/main/java/com/intuit/wasabi/assignmentobjects/Assignment.java
+++ b/modules/assignment-objects/src/main/java/com/intuit/wasabi/assignmentobjects/Assignment.java
@@ -284,6 +284,9 @@ public class Assignment {
         EXPERIMENT_IN_DRAFT_STATE(false),
         EXPERIMENT_PAUSED(false),
         NO_PROFILE_MATCH(false),
+        // EXPERIMENT_EXPIRED is no longer used since the experiment is moved to PAUSED upon expiring.
+        // The status is kept for the historical data.
+        EXPERIMENT_EXPIRED(false),
         ASSIGNMENT_FAILED(false),
 
         EXISTING_ASSIGNMENT(true),
--- a/modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
+++ b/modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
@@ -618,21 +618,20 @@ public class AssignmentsImpl implements Assignments {
 
         Assignment assignment = getAssignment(experimentID, userID, context, userAssignments, bucketList);
 
-        // Case: Experiment expires -> existing Assignments are returned and new Assignments
-        // are returned as null assignments
         if (currentTime > experiment.getEndTime().getTime()) {
-
             // the experiment moves to the PAUSED state if it was RUNNING before
-            if (Experiment.State.RUNNING == experiment.getState()) {
-                experimentUtil.updateExperimentState(experiment, Experiment.State.PAUSED);
-                metadataCache.refresh();
-            }
-
-            if (assignment != null && Assignment.Status.EXISTING_ASSIGNMENT == assignment.getStatus()) {
-                return assignment;
-            } else {
-                return nullAssignment(userID, applicationName, experimentID,
-                        Assignment.Status.EXPERIMENT_PAUSED);
+            if (Experiment.State.RUNNING.equals(experiment.getState())) {
+                // check if the information matches the one from the database
+                Experiment experimentInfoFromDB = experimentUtil.getExperiment(experiment.getID());
+                if (Experiment.State.RUNNING.equals(experimentInfoFromDB.getState())) {
+                    experimentUtil.updateExperimentState(experiment, Experiment.State.PAUSED);
+                    metadataCache.refresh();
+                    // update the object such that later checks return the right state
+                    experiment.setState(Experiment.State.PAUSED);
+                } else {
+                    // if they are not matching the database state is the new ground truth
+                    experiment.setState(experimentInfoFromDB.getState());
+                }
             }
         }
 
--- a/modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
+++ b/modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
@@ -515,10 +515,8 @@ public class AssignmentsImpl implements Assignments {
                     }
 
                     // Add the assignment to the global list of userAssignments of the user
-                    if (Experiment.State.PAUSED != experiment.getState()) { // the Experiment is set to PAUSED in the case of a reached endDate
-                        userAssignments.put(experiment.getID(), experiment.getLabel(),
-                                assignment.getBucketLabel() != null ? assignment.getBucketLabel().toString() : "null");
-                    }
+                    userAssignments.put(experiment.getID(), experiment.getLabel(),
+                            assignment.getBucketLabel() != null ? assignment.getBucketLabel().toString() : "null");
 
                     assignmentPairs.add(new ImmutablePair<>(experimentMap.get(experiment.getID()), assignment));
 
--- a/modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
+++ b/modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
@@ -17,7 +17,6 @@ package com.intuit.wasabi.assignment.impl;
 
 import com.datastax.driver.core.exceptions.ConnectionException;
 import com.google.common.collect.HashBasedTable;
-import com.google.common.collect.Sets;
 import com.google.common.collect.Table;
 import com.google.inject.Inject;
 import com.google.inject.name.Named;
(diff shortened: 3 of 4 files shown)


modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
```diff
@@ -734,7 +733,7 @@ public Response getMetadataCacheDetails() {
         }
 
         // Only include `assignment` property if there is a definitive assignment, either to a bucket or not
-        if (assignment.getStatus() != EXPERIMENT_EXPIRED) {
+        if (assignment.getStatus().isDefinitiveAssignment()) {
```
bargenilesh: @AndreaSuckro can you check with @calvinwma  about this conditional change as this change is not backward compatible especially for PAUSED experiments...

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -497,12 +486,12 @@ public Assignment putAssignment(User.ID userID, Application.Name applicationName
                     }
 
                     // Add the assignment to the global list of userAssignments of the user
-                    if (assignment.getStatus() != Assignment.Status.EXPERIMENT_EXPIRED) {
+                    if (Experiment.State.PAUSED != experiment.getState()) { // the Experiment is set to PAUSED in the case of a reached endDate
```
bargenilesh: @AndreaSuckro This change is also NOT backward compatible, Please check with @calvinwma ... @calvinwma A suggestion, to keep it backward compatible; now we may completely remove this IF condition as we are getting rid of EXPERIMENT_EXPIRED state and we used to add entry in the 'userAssignments' map even in case of Experiment.State.PAUSED

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -600,8 +589,11 @@ protected Assignment getAssignment(User.ID userID, Application.Name applicationN
         }
 
         if (currentTime > experiment.getEndTime().getTime()) {
+            // the experiment moves to the PAUSED state if the endtime is reached
+            experiment.setState(Experiment.State.PAUSED);
+            repository.updateExperiment(experiment);
```
bargenilesh: @AndreaSuckro This code is potentially going to update the state of experiment to be PAUSED and so send multiple emails, because of two reasons #1 Race condition in the multiple threads and #2 Metadata cache refreshes at the frequency of 5mins so assignment flow will not see the state change and so keep coming here... Please refer to the isAssignmentValidForRapidExperiment() implementation for the reference...

AndreaSuckro:
@bargenilesh I think I addressed #2, but am still a bit unsure about #1 - let's set up some time with @calvinwma to sort this out?

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
bargenilesh: @AndreaSuckro I take back my comment, I think technically this IF condition is NOT backward compatible but practically there are no bucket labels in case of NON-definitive assignment so this change looks good 👍

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
bargenilesh: @AndreaSuckro @calvinwma more description:  As per current business logic, user is attempted to assigned to the higher priority experiment, even if that higher priority experiment is PAUSED (assignment could be EXISTING_ASSIGNMENT or EXPERIMENT_PAUSED), user should NOT be attempted to be assigned to any other lower priority experiment...

This business rule will break if above  <IF> condition is added...

modules/api/src/main/java/com/intuit/wasabi/api/AssignmentsResource.java
abhishekjain88: 👍 @bargenilesh @AndreaSuckro

modules/assignment-objects/src/main/java/com/intuit/wasabi/assignmentobjects/Assignment.java
```diff
@@ -284,6 +284,8 @@ public String toString() {
         EXPERIMENT_IN_DRAFT_STATE(false),
         EXPERIMENT_PAUSED(false),
         NO_PROFILE_MATCH(false),
+        // EXPERIMENT_EXPIRED is no longer used since the experiment is moved to PAUSED upon expiring.
+        // The status is kept for the historical data.
```
abhishekjain88: 👍 for adding this comment @AndreaSuckro . Can you also mention that this historical data is only in downstream systems i.e. Hive/Vertica and not in any of the application DBs - Cassandra & MySQL for more clarity?

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -516,12 +515,12 @@ public Assignment putAssignment(User.ID userID, Application.Name applicationName
...
 
                     // Add the assignment to the global list of userAssignments of the user
-                    if (assignment.getStatus() != Assignment.Status.EXPERIMENT_EXPIRED) {
+                    if (Experiment.State.PAUSED != experiment.getState()) { // the Experiment is set to PAUSED in the case of a reached endDate
                         userAssignments.put(experiment.getID(), experiment.getLabel(),
                                 assignment.getBucketLabel() != null ? assignment.getBucketLabel().toString() : "null");
                     }
 
-                    assignmentPairs.add(new ImmutablePair<Experiment, Assignment>(experimentMap.get(experiment.getID()), assignment));
+                    assignmentPairs.add(new ImmutablePair<>(experimentMap.get(experiment.getID()), assignment));
```
abhishekjain88: 👍 @AndreaSuckro for removing redundant type information here.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -478,9 +477,9 @@ public Assignment putAssignment(User.ID userID, Application.Name applicationName
         //Prepopulate bucket assignment counts for rapid experiments in running state in asynchronous mode
         List<Experiment.ID> experimentIds =
                 appPriorities.getPrioritizedExperiments()
-                .stream()
-                .filter(experiment -> (null != experiment.getIsRapidExperiment() && experiment.getIsRapidExperiment())
-                        && experiment.getState().equals(Experiment.State.RUNNING))
+                        .stream()
```
abhishekjain88: 👍 @AndreaSuckro for correcting the formatting here.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -620,8 +619,12 @@ protected Assignment getAssignment(User.ID userID, Application.Name applicationN
         }
 
         if (currentTime > experiment.getEndTime().getTime()) {
+            // the experiment moves to the PAUSED state if the endtime is reached
+            experiment.setState(Experiment.State.PAUSED);
+            experimentUtil.updateExperimentState(experiment, Experiment.State.PAUSED);
```
abhishekjain88: @AndreaSuckro - Every time the `experimentUtil.updateExperimentState` method is called, an email is sent out to the admin of the experiment with the transition details of the state obtained as `experiment.getState().toString()` corresponds to previous state and the second parameter in the method corresponds to the new state. Refer to the `updateExperimentState()` method implementation.

So, there are 2 things that need to be fixed here:

1) Experiment state should only be updated if the experiment is in `RUNNING` state already. That way we would guarantee that the experiment state change is attempted only once and only one email gets sent out. To ensure guarantees around that, please refer to the last section of the `isAssignmentValidForRapidExperiment` method in the same class. You must ensure that you call the update and cache refresh event only once, as far as possible, avoiding all race conditions.

2) While calling the `updateExperimentState` method, please keep the state of the `experiment` object unchanged as it is used to refer to the existing state of the experiment in the state transition email.

modules/assignment/src/test/java/com/intuit/wasabi/assignment/impl/AssignmentsImplTest.java
```diff
@@ -346,7 +346,8 @@ public void testGetSingleAssignmentNullAssignmentExperimentExpired() {
         Assignment result = assignmentsImpl.doSingleAssignment(user, appName, label, context, true, true, segmentationProfile, headers);
 
         //Verify result
-        assertThat(result.getStatus(), is(Assignment.Status.EXPERIMENT_EXPIRED));
+        assertThat(result.getStatus(), is(Assignment.Status.EXPERIMENT_PAUSED));
+        verify(experiment).setState(Experiment.State.PAUSED);
```
abhishekjain88: @AndreaSuckro You should also probably verify the call to update the experiment state using the `experimentUtil` object.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
calvinwma: @bargenilesh @AndreaSuckro We'll need to make sure that we don't break the priority rules.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -620,8 +619,12 @@ protected Assignment getAssignment(User.ID userID, Application.Name applicationN
...
+            metadataCache.refresh();
             return nullAssignment(userID, applicationName, experimentID,
```
bargenilesh: @AndreaSuckro Existing behavior: For PAUSED experiment, system returns EXISTING_ASSIGNMENT assignment if user was assigned to the experiment or returns EXPERIMENT_PAUSED assignment.

Can you think of maintaining the same behavior? 

If NULL-Assignment (EXPERIMENT_PAUSED) is returned from here then above behavior will break...

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
AndreaSuckro: I think I understand now - thank you for the further explanation 👍

AndreaSuckro:
Thank you @abhishekjain88 and @bargenilesh for the feedback!

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -619,12 +616,26 @@ protected Assignment getAssignment(User.ID userID, Application.Name applicationN
                     Assignment.Status.EXPERIMENT_NOT_STARTED);
         }
 
+        Assignment assignment = getAssignment(experimentID, userID, context, userAssignments, bucketList);
+
+        // Case: Experiment expires -> existing Assignments are returned and new Assignments
+        // are returned as null assignments
         if (currentTime > experiment.getEndTime().getTime()) {
```
bargenilesh: @AndreaSuckro Thx for trying to keep the existing behavior for Expired->PAUSED experiment BUT this change doesn't address the case where there was existing_assignment but bucket had been emptied... 

I would suggest that in this expire-check IF condition only handle the experiment STATE update logic and leave rest of the code flow as is...

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -619,12 +616,26 @@ protected Assignment getAssignment(User.ID userID, Application.Name applicationN
...
-            return nullAssignment(userID, applicationName, experimentID,
-                    Assignment.Status.EXPERIMENT_EXPIRED);
+
+            // the experiment moves to the PAUSED state if it was RUNNING before
+            if (Experiment.State.RUNNING == experiment.getState()) {
```
bargenilesh: @AndreaSuckro In addition to the in-memory experiment state check (should be RUNNING in memory), you need to add MySQL experiment state check (should be RUNNING in MySQL) to avoid multiple emails in a thread race condition - similar to the isAssignmentValidForRapidExperiment() method...

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -516,12 +515,10 @@ public Assignment putAssignment(User.ID userID, Application.Name applicationName
                     }
 
                     // Add the assignment to the global list of userAssignments of the user
-                    if (assignment.getStatus() != Assignment.Status.EXPERIMENT_EXPIRED) {
-                        userAssignments.put(experiment.getID(), experiment.getLabel(),
-                                assignment.getBucketLabel() != null ? assignment.getBucketLabel().toString() : "null");
-                    }
+                    userAssignments.put(experiment.getID(), experiment.getLabel(),
+                            assignment.getBucketLabel() != null ? assignment.getBucketLabel().toString() : "null");
```
bargenilesh: @AndreaSuckro 👍

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -619,12 +616,26 @@ protected Assignment getAssignment(User.ID userID, Application.Name applicationN
...
+        // Case: Experiment expires -> existing Assignments are returned and new Assignments
+        // are returned as null assignments
         if (currentTime > experiment.getEndTime().getTime()) {
-            return nullAssignment(userID, applicationName, experimentID,
-                    Assignment.Status.EXPERIMENT_EXPIRED);
+
+            // the experiment moves to the PAUSED state if it was RUNNING before
+            if (Experiment.State.RUNNING == experiment.getState()) {
+                experimentUtil.updateExperimentState(experiment, Experiment.State.PAUSED);
+                metadataCache.refresh();
```
bargenilesh: @AndreaSuckro 👍  for refreshing the cache here - this cache had to be refreshed when in-memory state for an experiment is RUNNING but experiment is expired.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
AndreaSuckro: The expire check now only updates the state of the experiment, but not only in the db but also on the object for the later changes. I also added an else condition for the case where the database holds a different state than the current experiment - not sure though what the best behavior is .. currently it sets the database state as the new state, but we could also throw an error.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
AndreaSuckro: 👍

AndreaSuckro:
@bargenilesh can you have another look now with the latest changes? Thanks!

AndreaSuckro:
Closing this since it is ancient :)

PR #267, opened 2017-07-19
Fixing problem that default search did not apply "notTerminated", whi…

longdogz:
…ch used to be the default.


gittify:
:shipit:

PR #268, opened 2017-07-19
Fix so we can create superadmins.  Problem was wrong field being used…

longdogz:
… from return from verify API.  On non-Intuit, that field had the username, too.  On Intuit, only the username field has the user name.


gittify:
:shipit:

PR #270, opened 2017-07-28
Made put assignment implementation in sync with the get/post single/batch assignments

bargenilesh:
Retrieve experiment based upon app-name & experiment-label by going through the application-experiments list as per their priorities (Similar to the get/post single/batch assignment flow)...

As application-experiments priority list only contains non-deleted experiment-ids, PUT flow is not going to find any deleted experiment for the assignment.

### DIFF
--- a/modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
+++ b/modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
@@ -874,7 +874,7 @@ public class AssignmentsImpl implements Assignments {
                 for (PrioritizedExperiment prioritizedExperiment : prioritizedExperimentListOptional.get().getPrioritizedExperiments()) {
                     if (experimentLabel.equals(prioritizedExperiment.getLabel())) {
                         //Upon match, get the complete experiment object from cache
-                        result = metadataCache.getExperimentById(prioritizedExperiment.getID()).orElseGet(null);
+                        result = metadataCache.getExperimentById(prioritizedExperiment.getID()).orElse(null);
                         break;
                     }
                 }
--- a/modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
+++ b/modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
@@ -866,7 +866,8 @@ public class AssignmentsImpl implements Assignments {
         Experiment result = null;
 
         if (metadataCacheEnabled) {
-            //First fetch experiment list sorted by priorities (only contains non-deleted experiment-ids)
+            //First fetch experiment list sorted by priorities (contains non-deleted experiment-ids).
+            //This experiment-priority list is the source of truth for ALL assignment flows while looking up experiment...
             Optional<PrioritizedExperimentList> prioritizedExperimentListOptional = metadataCache.getPrioritizedExperimentListMap(applicationName);
             if (prioritizedExperimentListOptional.isPresent()) {
                 //Iterate as per experiment priority and look for the matching experiment by their label
--- a/modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
+++ b/modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
@@ -866,7 +866,7 @@ public class AssignmentsImpl implements Assignments {
         Experiment result = null;
 
         if (metadataCacheEnabled) {
-            //First fetch experiment list sorted by priorities
+            //First fetch experiment list sorted by priorities (only contains non-deleted experiment-ids)
             Optional<PrioritizedExperimentList> prioritizedExperimentListOptional = metadataCache.getPrioritizedExperimentListMap(applicationName);
             if (prioritizedExperimentListOptional.isPresent()) {
                 //Iterate as per experiment priority and look for the matching experiment by their label
@@ -874,6 +874,7 @@ public class AssignmentsImpl implements Assignments {
                     if (experimentLabel.equals(prioritizedExperiment.getLabel())) {
                         //Upon match, get the complete experiment object from cache
                         result = metadataCache.getExperimentById(prioritizedExperiment.getID()).orElseGet(null);
+                        break;
                     }
                 }
             }
--- a/modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
+++ b/modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
@@ -866,11 +866,12 @@ public class AssignmentsImpl implements Assignments {
         Experiment result = null;
 
         if (metadataCacheEnabled) {
-            List<Experiment> experiments = metadataCache.getExperimentsByAppName(applicationName);
-            for (Experiment exp : experiments) {
-                if (experimentLabel.equals(exp.getLabel())) {
-                    result = exp;
-                    break;
+            Optional<PrioritizedExperimentList> prioritizedExperimentListOptional = metadataCache.getPrioritizedExperimentListMap(applicationName);
+            if (prioritizedExperimentListOptional.isPresent()) {
+                for (PrioritizedExperiment prioritizedExperiment : prioritizedExperimentListOptional.get().getPrioritizedExperiments()) {
+                    if (experimentLabel.equals(prioritizedExperiment.getLabel())) {
+                        result = metadataCache.getExperimentById(prioritizedExperiment.getID()).orElseGet(null);
+                    }
                 }
             }
         } else {
--- a/modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
+++ b/modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
@@ -866,10 +866,13 @@ public class AssignmentsImpl implements Assignments {
         Experiment result = null;
 
         if (metadataCacheEnabled) {
+            //First fetch experiment list sorted by priorities
             Optional<PrioritizedExperimentList> prioritizedExperimentListOptional = metadataCache.getPrioritizedExperimentListMap(applicationName);
             if (prioritizedExperimentListOptional.isPresent()) {
+                //Iterate as per experiment priority and look for the matching experiment by their label
                 for (PrioritizedExperiment prioritizedExperiment : prioritizedExperimentListOptional.get().getPrioritizedExperiments()) {
                     if (experimentLabel.equals(prioritizedExperiment.getLabel())) {
+                        //Upon match, get the complete experiment object from cache
                         result = metadataCache.getExperimentById(prioritizedExperiment.getID()).orElseGet(null);
                     }
                 }
--- a/modules/assignment/src/test/java/com/intuit/wasabi/assignment/impl/AssignmentsImplTest.java
+++ b/modules/assignment/src/test/java/com/intuit/wasabi/assignment/impl/AssignmentsImplTest.java
@@ -1670,6 +1670,16 @@ public class AssignmentsImplTest {
         verify(metadataCache, times(1)).refresh();
     }
 
+    private static Experiment createMockExperiment(Experiment.ID experimentId, Experiment.Label exprimentLabel, Experiment.State experimentState) {
+        Experiment experiment = mock(Experiment.class, RETURNS_DEEP_STUBS);
+        when(experiment.getID()).thenReturn(experimentId);
+        when(experiment.getEndTime().getTime()).thenReturn(new Date().getTime() + 1000000L);
+        when(experiment.getLabel()).thenReturn(exprimentLabel);
+        when(experiment.getState()).thenReturn(experimentState);
+        when(experiment.getIsRapidExperiment()).thenReturn(false);
+        return experiment;
+    }
+
     @Test
     public void testExperimentFetchSameNameButDeleted() {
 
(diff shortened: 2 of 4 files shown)


modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -866,11 +866,15 @@ protected Experiment getExperiment(Application.Name applicationName, Experiment.
         Experiment result = null;
 
         if (metadataCacheEnabled) {
-            List<Experiment> experiments = metadataCache.getExperimentsByAppName(applicationName);
-            for (Experiment exp : experiments) {
-                if (experimentLabel.equals(exp.getLabel())) {
-                    result = exp;
-                    break;
+            //First fetch experiment list sorted by priorities
```
iizrailevsky: @bargenilesh Would it make sense to mention something re: the fact that prioritized experiments are never the deleted ones? Just to clarify why you need to "First fetch experiment list sorted by priorities"

modules/assignment/src/test/java/com/intuit/wasabi/assignment/impl/AssignmentsImplTest.java
```diff
@@ -1670,6 +1670,46 @@ public void testGetSingleAssignmentNullAssignmentRapidExperimentCapReached() {
...
+    @Test
+    public void testExperimentFetchSameNameButDeleted() {
+
+        //Create input objects
+        Application.Name appName = Application.Name.valueOf("Test");
+        Experiment.ID deleteExperimentId = Experiment.ID.newInstance(); //Experiment.ID.valueOf("");
+        Experiment.ID runningExperimentId = Experiment.ID.newInstance(); //Experiment.ID.valueOf("");
+        Experiment.Label commonExperimentLabel = Experiment.Label.valueOf("SameNamePresetTwice");
+
+        Experiment deleteExperiment = mock(Experiment.class, RETURNS_DEEP_STUBS);
```
iizrailevsky: @bargenilesh Nit: should this be called deletedExperiment (instead of deleteExperiment), to be contrasted with runningExperiment?

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -866,11 +866,15 @@ protected Experiment getExperiment(Application.Name applicationName, Experiment.
...
+            Optional<PrioritizedExperimentList> prioritizedExperimentListOptional = metadataCache.getPrioritizedExperimentListMap(applicationName);
+            if (prioritizedExperimentListOptional.isPresent()) {
+                //Iterate as per experiment priority and look for the matching experiment by their label
+                for (PrioritizedExperiment prioritizedExperiment : prioritizedExperimentListOptional.get().getPrioritizedExperiments()) {
+                    if (experimentLabel.equals(prioritizedExperiment.getLabel())) {
+                        //Upon match, get the complete experiment object from cache
+                        result = metadataCache.getExperimentById(prioritizedExperiment.getID()).orElseGet(null);
```
iizrailevsky: @bargenilesh Should there be a "break;" here after getting the result, similar to the previous implementation, to prevent further iteration?

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
bargenilesh: Done

modules/assignment/src/test/java/com/intuit/wasabi/assignment/impl/AssignmentsImplTest.java
bargenilesh: Done

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
bargenilesh: Nice catch, done...

gittify: @bargenilesh - looks good, one qn - what if prioritizedExperimentListOptional is not present

bargenilesh: @gittify In that case 'result' will be returned as NULL and which is handled in the putAssignment() method...

abhishekjain88: @bargenilesh Nit: If you wish, you can express the entire `for` loop more concisely with this Java 8 construct `result = prioritizedExperimentListOptional.get().getPrioritizedExperiments().stream().filter(prioritizedExperiment -> experimentLabel.equals(prioritizedExperiment.getLabel())).findFirst().map(prioritizedExperiment -> metadataCache.getExperimentById(prioritizedExperiment.getID()).orElse(null)).orElse(null); `

abhishekjain88: @bargenilesh Nit: Also, to @gittify 's point, I know this is true even for pre-existing code, but you could explicitly return an `Optional<Experiment>` from this method as it can potentially return a null value, very well serving the purpose of the `Optional` construct and making it more explicit that it can potentially return a nullable value.

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
```diff
@@ -866,11 +866,17 @@ protected Experiment getExperiment(Application.Name applicationName, Experiment.
...
+                        break;
+                    }
                 }
             }
         } else {
```
abhishekjain88: @bargenilesh Food for thought: The implementation for `getExperiment` made me think that why not also have a cache on top of the `experiment_label_index` table which is the source of this method call. It will help in de-cluttering a lot of iteration logic that we need to put together when getting the corresponding experiment from the priority list. Maybe something that we could add as a TODO/tech debt?

modules/assignment/src/test/java/com/intuit/wasabi/assignment/impl/AssignmentsImplTest.java
```diff
@@ -1670,6 +1670,46 @@ public void testGetSingleAssignmentNullAssignmentRapidExperimentCapReached() {
...
+        Experiment.Label commonExperimentLabel = Experiment.Label.valueOf("SameNamePresetTwice");
+
+        Experiment deletedExperiment = mock(Experiment.class, RETURNS_DEEP_STUBS);
+        when(deletedExperiment.getID()).thenReturn(deletedExperimentId);
+        when(deletedExperiment.getEndTime().getTime()).thenReturn(new Date().getTime() + 1000000L);
+        when(deletedExperiment.getLabel()).thenReturn(commonExperimentLabel);
+        when(deletedExperiment.getState()).thenReturn(Experiment.State.DELETED);
+        when(deletedExperiment.getIsRapidExperiment()).thenReturn(false);
+
+        Experiment runningExperiment = mock(Experiment.class, RETURNS_DEEP_STUBS);
```
abhishekjain88: @bargenilesh There is a lot of duplicate code between the creation + specification of the `deletedExperiment` and the `runningExperiment` objects, the only difference being the experiment state. Can we move this initialization to a separate private method which takes the state as a parameter and returns the created mock?

modules/assignment/src/test/java/com/intuit/wasabi/assignment/impl/AssignmentsImplTest.java
```diff
@@ -1670,6 +1670,46 @@ public void testGetSingleAssignmentNullAssignmentRapidExperimentCapReached() {
...
+        when(runningExperiment.getID()).thenReturn(runningExperimentId);
+        when(runningExperiment.getEndTime().getTime()).thenReturn(new Date().getTime() + 1000000L);
+        when(runningExperiment.getLabel()).thenReturn(commonExperimentLabel);
+        when(runningExperiment.getState()).thenReturn(Experiment.State.RUNNING);
+        when(runningExperiment.getIsRapidExperiment()).thenReturn(false);
+
+        PrioritizedExperimentList pExpList = new PrioritizedExperimentList();
+        pExpList.addPrioritizedExperiment(PrioritizedExperiment.from(runningExperiment, 1).build());
+        Optional<PrioritizedExperimentList> prioritizedExperimentListOptional = Optional.of(pExpList);
```
abhishekjain88: @bargenilesh Nit: Why declare this in a separate line? It is being used only once i.e. in the following line. Can we just create the optional there itself?

modules/assignment/src/test/java/com/intuit/wasabi/assignment/impl/AssignmentsImplTest.java
```diff
@@ -1670,6 +1670,46 @@ public void testGetSingleAssignmentNullAssignmentRapidExperimentCapReached() {
...
+        when(metadataCache.getPrioritizedExperimentListMap(appName)).thenReturn(prioritizedExperimentListOptional);
+        when(metadataCache.getExperimentById(runningExperimentId)).thenReturn(Optional.of(runningExperiment));
+        when(metadataCache.getExperimentById(deletedExperimentId)).thenReturn(Optional.of(deletedExperiment));
+
+        //Make actual call
+        Experiment returnedExperiment = assignmentsImpl.getExperiment(appName, commonExperimentLabel);
+
+        //Verify expected result
+        assertNotNull(returnedExperiment);
+        assertThat(returnedExperiment.getID(), is(runningExperimentId));
```
abhishekjain88: @bargenilesh Did not understand the purpose of the deleted experiment since you are not using it at all in the prioritized experiment list. I guess you would have added it to make it more evident that even if there are 2 experiments sharing the same name, the method would perform as expected. But the point is, as it is not being used/returned anywhere from the mocks used inside the method being tested, I think it can be safely dropped in its entirety from this test case.

modules/assignment/src/test/java/com/intuit/wasabi/assignment/impl/AssignmentsImplTest.java
```diff
@@ -1670,6 +1670,46 @@ public void testGetSingleAssignmentNullAssignmentRapidExperimentCapReached() {
...
+    }
```
abhishekjain88: @bargenilesh Do you also want to test for the scenario when the priority list does not contain any experiment being requested for and it returns null in that case?

Also, do we already have a test case for the `else` scenario where `metadateCache` is not enabled and we fallback on the actual fetch from the DB?

modules/assignment/src/test/java/com/intuit/wasabi/assignment/impl/AssignmentsImplTest.java
bargenilesh: deletedExperiment is being mocked to be returned by the 'metadataCache.getExperimentById(deletedExperimentId)' invocation... so if assignmentsImpl.getExperiment() uses it in future and mess-up the functionality it would get caught by this test case...

modules/assignment/src/test/java/com/intuit/wasabi/assignment/impl/AssignmentsImplTest.java
bargenilesh: @abhishekjain88 We don't have any tests where metadataCache is disabled...
@abhishekjain88 Added a test to test for returned NULL value..

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
bargenilesh: @abhishekjain88 Changed to use orElse(). For others points, it would be good to do those but as of now trying to keep as is, with just fixing the bug.

modules/assignment/src/test/java/com/intuit/wasabi/assignment/impl/AssignmentsImplTest.java
bargenilesh: @abhishekjain88 PrioritizedExperimentList don't provide a constructor to initialize with  PrioritizedExperiments...

modules/assignment/src/test/java/com/intuit/wasabi/assignment/impl/AssignmentsImplTest.java
bargenilesh: @abhishekjain88 Added a static method to create mockExperiment... Note: this method potentially could be used at many places but I am not changing any existing test cases...

modules/assignment/src/main/java/com/intuit/wasabi/assignment/impl/AssignmentsImpl.java
abhishekjain88: @bargenilesh 👍

modules/assignment/src/test/java/com/intuit/wasabi/assignment/impl/AssignmentsImplTest.java
abhishekjain88: @bargenilesh 👍

modules/assignment/src/test/java/com/intuit/wasabi/assignment/impl/AssignmentsImplTest.java
abhishekjain88: @bargenilesh I meant that the `Optional.of` initialization could be done directly in the next line where it is being used.

modules/assignment/src/test/java/com/intuit/wasabi/assignment/impl/AssignmentsImplTest.java
abhishekjain88: @bargenilesh Would it be good to add a test case to simulate metadataCache disabled as well?

modules/assignment/src/test/java/com/intuit/wasabi/assignment/impl/AssignmentsImplTest.java
abhishekjain88: @bargenilesh I still did not understand how this would help. Let's probably discuss it in person.

modules/assignment/src/test/java/com/intuit/wasabi/assignment/impl/AssignmentsImplTest.java
bargenilesh: @abhishekjain88 definitely, but I would think that as low priority and add in backlog, as metadataCache has been enabled in PROD for long time now and it has survived a tax-season too so it will be very unlikely that we would disable metadataCache.

modules/assignment/src/test/java/com/intuit/wasabi/assignment/impl/AssignmentsImplTest.java
bargenilesh: @abhishekjain88 sure...

modules/assignment/src/test/java/com/intuit/wasabi/assignment/impl/AssignmentsImplTest.java
abhishekjain88: @bargenilesh sounds good! Let's add a TODO. 👍

PR #272, opened 2017-08-01
Feature/ui sso login

longdogz:
These are changes in the UI code to enable support for alternate authentication methods, specifically, an SSO method where a user logs in and then is redirected to the site with secure cookies containing the authentication information, rather than having the user use the Sign In form.


modules/ui/app/views/SignIn.html
```diff
@@ -20,4 +20,8 @@
             <button id="btnSignin" class="blue">Sign In</button>
         </div>
     </form>
+    <div ng-if="dontShowForm">
+        <span ng-if="serverDown">We're sorry, but the Wasabi service is currently down.  Please <a href="{{redirectUrl}}">try again</a> later.</span>
```
abhishekjain88: We need to get rid of the redirect URL here, as discussed.

modules/ui/app/scripts/services/UtilitiesFactory.js
```diff
@@ -826,6 +826,67 @@ angular.module('wasabi.services').factory('UtilitiesFactory', ['Session', '$stat
                 return '';
             },
 
+            getPermissions: function(result, transitionToFirstPage) {
```
abhishekjain88: @longdogz Is this the change for superadmins which we are combining here?

Issue #293, opened 2017-11-06
Cannot find `wasabi-main`


leeDav:
Similar to: https://github.com/intuit/wasabi/issues/219 and https://github.com/intuit/wasabi/issues/283.

But this time I'm on Mac OS X (10.12.3).

Node: 8.9.0
npm: 5.5.1
nvm: 0.33.6
Java: 9.0.1

First time I try to install/bootstrap I get the below.
```sh
[INFO] ------------------------------------------------------------------------
[INFO] Skipping wasabi
[INFO] This project has been banned from the build due to previous failures.
[INFO] ------------------------------------------------------------------------
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO]
[INFO] wasabi ............................................. SUCCESS [  1.345 s]
[INFO] wasabi-experiment-objects .......................... SUCCESS [  4.604 s]
[INFO] wasabi-export ...................................... SUCCESS [  2.069 s]
[INFO] wasabi-assignment-objects .......................... SUCCESS [  2.507 s]
[INFO] wasabi-exceptions .................................. SUCCESS [  2.243 s]
[INFO] wasabi-analytics-objects ........................... SUCCESS [  3.533 s]
[INFO] wasabi-authentication-objects ...................... SUCCESS [  2.134 s]
[INFO] wasabi-eventlog .................................... SUCCESS [  2.854 s]
[INFO] wasabi-cassandra-datastax .......................... SUCCESS [  2.185 s]
[INFO] wasabi-authorization-objects ....................... SUCCESS [  1.971 s]
[INFO] wasabi-user-directory .............................. SUCCESS [  2.037 s]
[INFO] wasabi-feedback-objects ............................ SUCCESS [  1.661 s]
[INFO] wasabi-auditlog-objects ............................ SUCCESS [  1.915 s]
[INFO] wasabi-database .................................... SUCCESS [  2.415 s]
[INFO] wasabi-repository-datastax ......................... FAILURE [  0.929 s]
[INFO] wasabi-experiment .................................. SKIPPED
[INFO] wasabi-assignment .................................. SKIPPED
[INFO] wasabi-util ........................................ SKIPPED
[INFO] wasabi-analytics ................................... SKIPPED
[INFO] wasabi-auditlog .................................... SKIPPED
[INFO] wasabi-event-objects ............................... SKIPPED
[INFO] wasabi-event ....................................... SKIPPED
[INFO] wasabi-email ....................................... SKIPPED
[...]

PR #295, opened 2017-11-10
Fix build issues on Mac

bradphilips:
This is to fix build issues on the Mac regarding Issue #293. I'm not sure if this fixes the referenced issues, however, it does allow this to build and run on a Mac. The issue was with the lombok dependency generating the `wasabi-repository-datastax: Fatal error compiling: java.lang.NoClassDefFoundError: com/sun/tools/javac/file/BaseFileObject: com.sun.tools.javac.file.BaseFileObject` which was fixed in the lombok version 1.16.18. See issue [1315](https://github.com/rzwitserloot/lombok/issues/1315) in the lombok repository. Once this was added I had to also add some ignores to jdk.internal.reflect so PowerMock wouldn't try to stub these classes. Let me know if you have questions. Thanks!


s34gull:
I am now able to build wasabi after manually patching using this commit.

jrgifford:
Can confirm that this fixes it for me as well on 10.13.2

Issue #335, opened 2018-06-04
Missing Cassandra fields


mariusciziunas:
Hi dear Wasabi maintainers,

There are few fields in the code, but they are missing in Cassandra DB:
- experiment.source_url
- experiment.experiment_type
Are you aware of it? Do they need to be included or removed from both places?

Issue #265, opened 2017-07-19
Can't start on MacOS


llambda:
I did:

```
gco 1.0.20170717233424
```

followed by:

```
 bin/wasabi.sh start   
```

and am seeing:

```
Successfully built 9f3c7da36c97
wasabi: starting
4a14bc2da8b4ba73f61de270613b3c5100f29bad08cfb72897909228d7059a1d
chill'ax error: not started

usage: container.sh [options] [commands]

options:
  -v | --verify [ true | false ]         : verify installation configuration; default: false
  -m | --migration [ true | false ]      : refresh cassandra migration scripts; default: false
  -s | --sleep [ sleep-time ]            : sleep/wait time in seconds; default: 3
  -h | --help                            : help message

commands:
  start[:wasabi,cassandra,mysql,docker]  : start all, wasabi, cassandra, mysql, docker
  stop[:wasabi,cassandra,mysql,docker]   : stop all, wasabi, cassandra, mysql, docker
  console[:cassandra,mysql]              : console all, cassandra, mysql
  status                                 : status
  remove[:wasabi,cassandra,mysql]]       : remove all, wasabi, cassandra, mysql



```

brendanjryan:
@llambda is this all of the output? can you try starting each service individually? I'm betting that the mysql container is having issues coming up

symadept:
I'm having the same issue. 
Successfully tagged wasabi-main:mshai4-1525274647
Successfully tagged wasabi-main:latest
wasabi: starting
27b24b6714d55fef3f213a9ce275f8cafbed0f2116bd6d68dc7cc25554460793
chill'ax error: not started

usage: container.sh [options] [commands]

j33wang:
Having the same issue, anyone have more details on a workaround or how to go about debugging this?

It seems 'wasabi.sh status' is printing this out ("start:wasabi" calls "status") .

In 'wasabi.sh status', "wget -q --spider --tries=20 --waitretry=3 http://localhost:8080/api/v1/ping" isn't working.  

I'm seeing:
`
$ wget --spider http://localhost:8080/api/v1/ping
`
`
Spider mode enabled. Check if remote file exists.
--2018-05-08 19:39:52--  http://localhost:8080/api/v1/ping
Resolving localhost (localhost)... 127.0.0.1
Connecting to localhost (localhost)|127.0.0.1|:8080... connected.
HTTP request sent, awaiting response... Read error (Connection reset by peer) in headers.
Retrying.
`

[...]

PR #336, opened 2018-06-14
Update databases with missing fields

dine-j:
Following Issue #265 and #335, missing fields were added in Cassandra and sql.


robmorgan:
👍

nickcoutsos:
This should get reviewed and merged as soon as possible, I started to investigate using wasabi on a project and without this migration I wasn't able to get up and running with the latest revision (or any that isn't many months old).

It's particularly frustrating to new users because the only way to debug is to look through the container's docker logs and find a mention of the missing fields among around 100 other vague exceptions thrown in the cassandra driver.

PR #358, opened 2018-11-27
Fix for Race Condition when using Pagination API: https://github.com/…

hindog:
…intuit/wasabi/issues/356

Tested using python script in example.  After patch, script returns 0 errors.

### DIFF
--- a/modules/api/src/main/java/com/intuit/wasabi/api/ApiModule.java
+++ b/modules/api/src/main/java/com/intuit/wasabi/api/ApiModule.java
@@ -44,6 +44,7 @@ import com.intuit.wasabi.userdirectory.UserDirectoryModule;
 import org.slf4j.Logger;
 
 import java.util.Properties;
+import java.util.function.Supplier;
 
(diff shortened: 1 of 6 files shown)


hindog:
Concurrency issues appear to stem from using a shared, mutable `PaginationFilter` across API calls (same goes for `PaginationComparator` instances) .  For each API call, `filter.replaceFilter` is called, mutating the shared filter criteria, which will impact any other in-progress calls using that same filter instance.

The fix is to modify the `PaginationHelper` instance to get injected with `Supplier<PaginationHelper>` and `Supplier<PaginationComparator>` so that each invocation of `paginate` will get a fresh instance of these classes, eliminating the shared filter/sort state across requests. 

Filters/Comparators impacted, and subsequently fixed by this patch:

```
PaginationFilter<AuditLogEntry>
PaginationComparator<AuditLogEntry>
PaginationFilter<Experiment>
PaginationComparator<Experiment>
PaginationFilter<ExperimentDetail>
PaginationComparator<ExperimentDetail>
```

PR #417, opened 2021-03-17
Fetch active experiment id for given experiment label

riteshmodi:


modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/impl/CassandraExperimentRepository.java
```diff
@@ -333,16 +333,24 @@ protected Experiment internalGetExperiment(Application.Name appName,
         Experiment experiment = null;
 
         try {
-            ExperimentByAppNameLabel experimentAppLabel = experimentLabelIndexAccessor
-                    .getExperimentBy(appName.toString(),
-                            experimentLabel.toString()).one();
```
riteshmodi, line 338: If I remember correctly, then value fetched in this method are cached by wasabi. Did you notice `one()`? That means, it will fetch only 1 experiment label if there are 2.

modules/repository-datastax/src/main/java/com/intuit/wasabi/repository/cassandra/impl/CassandraExperimentRepository.java
```diff
@@ -333,16 +333,24 @@ protected Experiment internalGetExperiment(Application.Name appName,
...
+            Result<ExperimentByAppNameLabel> result = experimentLabelIndexAccessor
+                    .getExperimentBy(appName.toString(),
+                            experimentLabel.toString());
+            for(ExperimentByAppNameLabel experimentAppLabel: result.all()){
+                if (experimentAppLabel != null) {
+                    // Need to iterate here as well
+                    com.intuit.wasabi.repository.cassandra.pojo.Experiment experimentPojo = experimentAccessor
+                            .getExperimentById(experimentAppLabel.getId()).one();
+                    experiment = ExperimentHelper.makeExperiment(experimentPojo);
+                    if(experiment.getState().isActiveState()){
```
riteshmodi, line 346: This will return only the experiment which is active

